PR #9, opened 2022-01-06
revenue sharing POC

mortelli:
PR to merge work between myself and @antomor for the revenue sharing POC.


src/RelayClient.ts
```diff
@@ -633,6 +643,39 @@ export class RelayClient {
                 relayInfo
             )} transaction: ${JSON.stringify(transactionDetails)}`
         );
+        if ([undefined, null].includes(transactionDetails.tokenAmount)) {
+            log.debug('Calculating maxPossibleGas...');
+
+            const txDetailsClone: EnvelopingTransactionDetails = {
+                ...transactionDetails,
+                // a dump value, just to allow the correct estimation, including the token transfer
+                tokenAmount: web3.utils.toWei('1').toString()
```
mortelli, line 652: a couple of questions:
1. does this fail without a `tokenAmount` set?
2. does it make a difference which value we use? would `0` work?

src/RelayClient.ts
```diff
@@ -956,3 +1003,21 @@ export function _dumpRelayingResult(relayingResult: RelayingResult): string {
...
+function getTRifWei(costInWei: BN) {
+    const tRifPriceInRBTC = 0.000005739;
+    const ritTokenDecimals = 18;
+
+    const costInRBTC = web3.utils.fromWei(costInWei.toString());
+    const costInTrif = parseFloat(costInRBTC) / tRifPriceInRBTC;
+    const costInTrifFixed = costInTrif.toFixed(ritTokenDecimals);
+    const costInTrifAsWei = web3.utils.toWei(costInTrifFixed);
+    return costInTrifAsWei;
+}
```
mortelli, line 1023: i know it is temporary and hard-coded, but i would move this to a separate file for clearness' sake

mortelli (commented):
thanks

please don't forget to update the dist after applying changes (if any)

src/RelayClient.ts
antomor: No, it doesn't fail, but it's required in order to estimate the complete workflow. Without the `tokenAmount` set, it will estimate the flow where the `tokenAmount` is not set. For instance, have a look at [the SmartWallet](https://github.com/anarancio/rif-relay-contracts/blob/master/contracts/smartwallet/SmartWallet.sol#L113) contract

mortelli: got it. makes perfect sense. thanks!

src/Conversions.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * For the moment we hard-coded the trif-rbtc conversion (extracted from the original dapp)
+ */
+export function costInWei(maxPossibleGas: BN, gasPrice: BN) {
+    return maxPossibleGas.mul(gasPrice);
+}
+
```
mortelli, line 7:
```suggestion
/**
 * TODO: Hard-coded values: for testing purposes only!
 This is specific for the tRIF token.
 */
```

src/RelayClient.ts
mortelli, line 1023: duplicated code right?

src/Conversions.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * For the moment we hard-coded the trif-rbtc conversion (extracted from the original dapp)
+ */
+export function costInWei(maxPossibleGas: BN, gasPrice: BN) {
```
mortelli, line 4: shouldn't `gas` be somewhere in the name of this function? not a blocker

src/RelayClient.ts
```diff
@@ -624,6 +624,16 @@ export class RelayClient {
         return `0x${gasPrice.toString(16)}`;
     }
 
+    async getMaxPossibleGas(
```
mortelli, line 627: are we using this somewhere?

mortelli (commented):
some comments

src/RelayClient.ts
antomor: My fault I didn't push this change

src/RelayClient.ts
antomor: I don't think so, good catcha!

PR #11, opened 2022-02-11
PP-94: Fix issues arisen from integration tests

antomor:
## What

- Check if the collector contract is defined during the estimation of the token transfer gas
- Remove automatic token amount calculation when it's undefined or null

## Why

- We don't need to forcefully calculate the `tokenAmount` parameter if it's not set by the user.
- On token transfer gas estimation, check if the collector gas is undefined/null or if it is equal to the zero address to decide the token receiver


package.json
```diff
@@ -18,8 +18,8 @@
   "author": "Relaying Services Team",
   "license": "MIT",
   "dependencies": {
-    "@rsksmart/rif-relay-common": "https://github.com/infuy/rif-relay-common#PP-94/revenue-sharing-model",
-    "@rsksmart/rif-relay-contracts": "https://github.com/anarancio/rif-relay-contracts#PP-94/revenue-sharing-model",
+    "@rsksmart/rif-relay-common": "https://github.com/antomor/rif-relay-common#PP-94/mandatory-collector-address",
+    "@rsksmart/rif-relay-contracts": "https://github.com/antomor/rif-relay-contracts#PP-94/fix-truffle-typings-patch",
```
antomor, line 22: We need to revert this change after merging the PRs:
- [rif-relay-common](https://github.com/infuy/rif-relay-common/pull/9)
- [rif-relay-contracts](https://github.com/anarancio/rif-relay-contracts/pull/21)

mortelli (commented):
the token receiver change seems good :+1: 

just 1 question regarding the estimation change: if the token amount is undefined, what's going to happen now that we have removed the forced conversions?

antomor:
> just 1 question regarding the estimation change: if the token amount is undefined, what's going to happen now that we have removed the forced conversions?

If the token amount is undefined/null, [0 is used](https://github.com/antomor/rif-relay-client/blob/98aba9ae8be8e5f5d362c961aa634183f0b98f6e/src/RelayClient.ts#L881) for both `tokenAmount` and `tokenGas`.

mortelli (approved):

PR #13, opened 2022-05-21
feature(devops): adds ci - runs checks on pushes

david-iov:


.github/workflows/main.yml
```diff
@@ -0,0 +1,26 @@
...
+                  node-version: '16.x'
+
+            - name: Install dependencies
+              run: npm install
+
+            - name: Check Linter bugs
+              run: npm run lint
+
+            - name: Fix Linter bugs
+              run: npm run lint:fix
```
antomor, line 20: I think that this task will never be executed because it's necessary when the previous one fails, so it's intended to be used locally by devs

.github/workflows/main.yml
```diff
@@ -0,0 +1,26 @@
...
+
+            - name: Check Codestyles errors
+              run: npm run prettier
+
+            - name: Fix Codestyles bugs
+              run: npm run prettier:fix
```
antomor, line 26: I think that this task will never be executed because it's necessary when the previous one fails, so it's intended to be used locally by devs

antomor (changes requested):
I'd remove the `*:fix` tasks because they are executed when the check tasks fail

antomor (approved):
LGTM

PR #15, opened 2022-05-30
feature: extracts code to function that calculates internal cost

ironFe93:
## What

Extracted a portion of code that is used to calculate the internal cost of a transaction, put it into a public function inside relayClient.

## Why

Need it exposed in relay client, to be used in rif-relay-sdk. 

## Refs


antomor (approved):
LGTM

PR #18, opened 2022-06-16
PP-137/Unnecessary domainSeparator field in Request

AndresQuijano:
## What

- Modifiy some types and functions in order to avoid passing the domain separator as part of the request.

## Why

- This extra and unnecessary field makes the operation more expensive.

## Refs

- https://rsklabs.atlassian.net/browse/PP-137
- https://github.com/rsksmart/rif-relay/issues/181


antomor:
Close in favor of #19

PR #20, opened 2022-06-28
feature/PP-171

franciscotobar:
## What

- Created a transaction history modal to be able to verify the transactions for each smart wallet

## Why

- With the current implementation, each time the users try to execute a transaction, they need to wait until it is confirmed (or rejected). To avoid this, we could provide the user with the link to the explorer.

## Refs

- [Jira issue](https://rsklabs.atlassian.net/browse/PP-171)
- https://github.com/rsksmart/rif-relay-common/pull/23

### DIFF
--- a/src/RelayProvider.ts
+++ b/src/RelayProvider.ts
@@ -2,7 +2,7 @@
 import abiDecoder from 'abi-decoder';
 import log from 'loglevel';
 import { JsonRpcPayload, JsonRpcResponse } from 'web3-core-helpers';
-import { HttpProvider } from 'web3-core';
+import { HttpProvider, TransactionReceipt } from 'web3-core';
 import { IRelayHub, IWalletFactory } from '@rsksmart/rif-relay-contracts';
 import {
     _dumpRelayingResult,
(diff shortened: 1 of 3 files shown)


src/RelayProvider.ts
```diff
@@ -366,7 +366,8 @@ export default class RelayProvider implements HttpProvider {
 
     _ethSendTransaction(
         payload: JsonRpcPayload,
-        callback: JsonRpcCallback
+        callback: JsonRpcCallback,
+        nonBlock?: boolean
```
antomor, line 370: I don't think we should add this parameter here because it may be ignored when this method is called by the `send` method. The `RelayProvider` implements the `HttpProvider` that extends the `HttpProviderBase`. There is a `send` method with the following signature:
```typescript
 send(
        payload: JsonRpcPayload,
        callback: (error: Error | null, result?: JsonRpcResponse) => void
    ): void;
``` 

This method is overwritten by the RelayProvider, so I think we may need to add the `nonBlock` parameter in the payload directly to be compatible with the interface. Furthermore, we may need to change the [SDK](https://github.com/rsksmart/rif-relay-sdk/blob/master/src/sdk.ts#L302) as well to receive this parameter and/or to include it in the payload of the request.

**NIT**: I would change the name of the parameter to `waitForTransactionReceipt`, just to clarify what it's used for.

antomor (changes requested):
@franciscotobar Thank you for the changes.
I'm afraid that we are breaking the compatibility with the HttpProvider interface.

src/RelayProvider.ts
franciscotobar: I just did the changes :)

antomor (approved):
LGTM

PR #23, opened 2022-07-21
feature/PP-204

franciscotobar:
## What

- Validation for smart wallet address to verify if it belongs to the EOA and its deployed

## Why

- To be able to import a smart wallet

## Refs

- [Jira issue](https://rsklabs.atlassian.net/browse/PP-204)
- https://github.com/rsksmart/rif-relay-common/pull/27


test/RelayClient.test.ts
```diff
@@ -0,0 +1,143 @@
...
+            await relayClient.validateSmartWallet(fakeTransactionDetails);
+            assert.isTrue(
+                spyCall.calledOnceWithExactly(fakeTransactionDetails),
+                'Was not called once'
+            );
+        });
+
+        it('should use contractInteractor.getSenderNonce once', async () => {
+            await relayClient.validateSmartWallet(fakeTransactionDetails);
+            expect(fakeContractInteractor.getSenderNonce).to.have.been.called;
```
antomor, line 123: we're not checking how many times the method has been called

test/RelayClient.test.ts
```diff
@@ -0,0 +1,143 @@
...
+        });
+
+        it('should use accountManager.sign once', async () => {
+            await relayClient.validateSmartWallet(fakeTransactionDetails);
+            expect(fakeAccountManager.sign).to.have.been.called;
```
antomor, line 128: we're not checking how many times the method has been called

test/RelayClient.test.ts
```diff
@@ -0,0 +1,143 @@
...
+        });
+
+        it('should use contractInteractor.verifyForwarder once', async () => {
+            await relayClient.validateSmartWallet(fakeTransactionDetails);
+            expect(fakeContractInteractor.verifyForwarder).to.have.been.called;
```
antomor, line 133: we're not checking how many times the method has been called

antomor (changes requested):
@franciscotobar Apart from the comment, I think that here we are testing for the method implementation. I mean, if we test that all inner objects are called, we'll probably need to change all those tests if we change the method implementation. IMHO we should at least include all these checks in one single test, because after all, we're testing for its flow, so we need all of them to be checked. So in summary, I think we could check all the method calls in one single test.
However, if someone from @rsksmart/payments has different opinions, I'd be glad to hear.

AndresQuijano (approved):
LGTM

antomor (approved):
LGTM

PR #24, opened 2022-07-26
fix/PP-171

franciscotobar:
## What

- Fixing feature/PP-171

## Why

- Integration testing were not being successful

### DIFF
--- a/src/RelayProvider.ts
+++ b/src/RelayProvider.ts
@@ -19,6 +19,7 @@ import { AccountKeypair } from './AccountManager';
 import { RelayEvent } from './RelayEvents';
 import { Address } from './types/Aliases';
 import { toBN, toChecksumAddress, toHex } from 'web3-utils';
+import { Transaction } from 'ethereumjs-tx';
 
 // @ts-ignore
 abiDecoder.addABI(IRelayHub.abi);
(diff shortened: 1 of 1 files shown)


src/RelayProvider.ts
```diff
@@ -19,6 +19,7 @@ import { AccountKeypair } from './AccountManager';
 import { RelayEvent } from './RelayEvents';
 import { Address } from './types/Aliases';
 import { toBN, toChecksumAddress, toHex } from 'web3-utils';
+import { Transaction } from 'ethereumjs-tx';
```
antomor, line 22: Could you please remove this if not necessary?

src/RelayProvider.ts
```diff
@@ -237,6 +242,73 @@ export default class RelayProvider implements HttpProvider {
         return this.executeRelayTransaction(transactionDetails);
     }
 
+    async processTransactionReceipt(
```
antomor, line 245: probably this method should be private, shouldn't it?

antomor (changes requested):

src/RelayProvider.ts
franciscotobar: done

src/RelayProvider.ts
franciscotobar: done

antomor (approved):
LGTM

PR #28, opened 2022-08-17
feature/PP-288

franciscotobar:
## What

- Integration of price feeder for RBTC-RIF

## Why

- Needed for revenue sharing validation process

## Refs
 - [PR #288 (Implement a token-RBTC conversion)](https://rsklabs.atlassian.net/browse/PP-288)


src/RelayPricer.ts
```diff
@@ -0,0 +1,34 @@
...
+        if (typeof sourceApi === 'string' || sourceApi instanceof String) {
+            switch (sourceApi as string) {
+                case 'coinbase':
+                    this.sourceApi = new CoinBase();
+                    break;
+                default:
+                    throw new Error('Source api not provided');
+            }
+        } else {
+            this.sourceApi = sourceApi as SourceApi;
```
ironFe93, line 17: Does this mean that we can send in any given API?

src/RelayPricer.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    async getPrice(
+        pair1: string,
+        pair2: string,
+        intermediaryPair?: string
+    ): Promise<number> {
+        const intermediary = intermediaryPair ? intermediaryPair : 'USD';
+        const price1 = await this.sourceApi.query(pair1, intermediary);
+        const price2 = await this.sourceApi.query(pair2, intermediary);
+        if (price1 === 0 || price2 === 0) {
```
ironFe93, line 29: can the price ever be zero? what happens if I send in a currency that doesn't exist? also maybe instead of pair1 and pair2, we might call them baseCurrency and counterCurrency, or sourceCurrency and targetCurrency.

src/api/CoinBase.ts
```diff
@@ -0,0 +1,27 @@
+import { CoinBaseResponse, SourceApi } from '../types/RelayPricer';
+import fetch, { Response } from 'node-fetch';
+
+export default class CoinBase implements SourceApi {
+    // we should get this address from config file or enviroments
+    private readonly URL: string = 'https://api.coinbase.com/v2/exchange-rates';
+
+    async query(input: string, output: string): Promise<number> {
+        output = output.toUpperCase();
```
ironFe93, line 9: better to create another variable, const upperCaseOutput, instead of mutating the same variable.

src/api/CoinBase.ts
```diff
@@ -0,0 +1,27 @@
...
+        const response: Response = await fetch(
+            `${this.URL}?currency=${input}`,
+            {
+                method: 'get'
+            }
+        );
+
+        if (response.status === 200) {
+            const result: CoinBaseResponse =
+                (await response.json()) as CoinBaseResponse;
```
ironFe93, line 19: how about 
```suggestion
               const result: CoinBaseResponse = await response.json();
```

ironFe93 (approved):
I think its good.

src/RelayPricer.ts
franciscotobar: Thats the idea, to give more flexibility.

src/RelayPricer.ts
franciscotobar: Since we can use any given API, there is a chance that someone return zero if the token does not exists. At this point if the token does not exist, in the CoinBase implementation I return an error.

src/api/CoinBase.ts
franciscotobar: I wasn't sure about this. I'll be updating it.

src/api/CoinBase.ts
franciscotobar: Let me verify, I remember that I had issue with typescript and was not allowing me to do that.

src/api/CoinBase.ts
franciscotobar: done

src/api/CoinBase.ts
franciscotobar: done

src/RelayPricer.ts
franciscotobar: done

src/RelayPricer.ts
```diff
@@ -0,0 +1,42 @@
+import CoinBase from './api/CoinBase';
+import { SourceApi } from './types/RelayPricer';
+
+export default class RelayPricer {
+    private readonly sourceApi: SourceApi;
+
+    constructor(sourceApi: SourceApi | string) {
```
antomor, line 7: **NIT**:
I'd simplify the constructor to receive argument of type `SourceApi`. IMHO the creation of the `SourceAPI` object should be done outside this class.

src/RelayPricer.ts
```diff
@@ -0,0 +1,42 @@
...
+            ? intermediateCurrency
+            : 'USD';
+        const sourcePrice = await this.sourceApi.query(
+            sourceCurrency,
+            intermediary
+        );
+        const targetPrice = await this.sourceApi.query(
+            targetCurrency,
+            intermediary
+        );
```
antomor, line 36: I think that those to calls can be performed in parallel

test/RelayPricer.test.ts
```diff
@@ -0,0 +1,86 @@
...
+            );
+        });
+
+        it('should return source currency conversion price', async () => {
+            fakeSourceApi.query
+                .withArgs(sourceCurrency, intermediateCurrency)
+                .returns(Promise.resolve(fakePrice1));
+            fakeSourceApi.query
+                .withArgs(targetCurrency, intermediateCurrency)
+                .returns(Promise.resolve(fakePrice2));
```
antomor, line 39: it looks like this code is duplicated on line 45 as well. We may think about creating a function for doing this.

test/RelayPricer.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    const fakePriceConversion1 = 300892.2291221775;
+    const fakePriceConversion2 = 0.0000033234490731694805;
+    const sourceCurrency = 'RBTC';
+    const targetCurrency = 'RIF';
+    const intermediateCurrency = 'USD';
+
+    describe('getPrice', async () => {
+        beforeEach(() => {
+            fakeSourceApi = stubInterface<SourceApi>();
+            pricer = new RelayPricer(fakeSourceApi);
```
antomor, line 24: I think that we are not testing for the constructor code.

test/RelayPricer.test.ts
```diff
@@ -0,0 +1,86 @@
...
+                .withArgs(sourceCurrency, intermediateCurrency)
+                .returns(Promise.resolve(0));
+            fakeSourceApi.query
+                .withArgs(targetCurrency, intermediateCurrency)
+                .returns(Promise.resolve(fakePrice2));
+            await assert.isRejected(
+                pricer.getPrice(sourceCurrency, targetCurrency),
+                'price cannot be zero'
+            );
+        });
```
antomor, line 84: I think here we are testing just one case where the `sourceCurrency` is 0, but what if the `targetCurrency` is 0?

test/api/CoinBase.test.ts
```diff
@@ -0,0 +1,60 @@
...
+        it('should fail if token/coin does not exist', async () => {
+            fakeResponse = new Response(undefined, { status: 400 });
+            fakeFetch = stub(fetchModule, 'default').returns(
+                Promise.resolve(fakeResponse)
+            );
+            await assert.isRejected(
+                coinBase.query('NA', output),
+                'Input not available'
+            );
+        });
```
antomor, line 58: I think we are not testing here the successful path 😅

antomor (changes requested):
@franciscotobar thank you for your changes, I like the solution applied 👍🏼 
Just a question: are you planning to open a different PR in the server where we use it? Or were you considering to do this in a different task?

src/RelayPricer.ts
franciscotobar: We could that change, but I was thinking more about having the implementations on the rif-relay-client, and not exporting the SourceApi

test/RelayPricer.test.ts
franciscotobar: Let me create a function

test/api/CoinBase.test.ts
franciscotobar: updated

test/RelayPricer.test.ts
franciscotobar: updated

test/RelayPricer.test.ts
franciscotobar: I added an extra comment where it was tested

test/RelayPricer.test.ts
franciscotobar: updateds

src/RelayPricer.ts
franciscotobar: updated

franciscotobar:
After this PR is approved, I think that I'll cherry pick this commits to the revenue sharing feature and there, create the PR for the server, since in the revenue sharing is where is used, isn't it?

src/RelayPricer.ts
franciscotobar: updated

test/RelayPricer.test.ts
franciscotobar: updated

src/RelayPricer.ts
```diff
@@ -0,0 +1,28 @@
+import { SourceApi } from './types/RelayPricer';
+
+export default class RelayPricer {
+    private readonly sourceApi: SourceApi;
+
+    constructor(sourceApi: SourceApi) {
+        this.sourceApi = sourceApi;
+    }
+
```
AndresQuijano, line 9: What if I pass a null/undefined sourceApi to this constructor?

src/RelayPricer.ts
```diff
@@ -0,0 +1,28 @@
...
+            ? intermediateCurrency
+            : 'USD';
+
+        const [sourcePrice, targetPrice] = await Promise.all([
+            this.sourceApi.query(sourceCurrency, intermediary),
+            this.sourceApi.query(targetCurrency, intermediary)
+        ]);
+        if (sourcePrice === 0 || targetPrice === 0) {
+            throw new Error('price cannot be zero');
+        }
```
AndresQuijano, line 25: This message is being returned to the caller but the caller is not the source of this 0. I think this message can be improved.

src/api/CoinBase.ts
```diff
@@ -0,0 +1,26 @@
...
+    async query(input: string, output: string): Promise<number> {
+        const upperCaseOutput = output.toUpperCase();
+        const response: Response = await fetch(
+            `${this.URL}?currency=${input}`,
+            {
+                method: 'get'
+            }
+        );
+
+        if (response.status === 200) {
```
AndresQuijano, line 17: Please define a descriptive constant to use instead of 200

src/RelayPricer.ts
antomor: Did you mean that you wouldn't export the SourceApi outside the rif-relay-client library? If so, we could build and export a factory that performs what you were doing in the constructor before.

src/RelayPricer.ts
antomor: Furthermore, to store the error message we could use a constant which will be reused in the tests

antomor:
> After this PR is approved, I think that I'll cherry pick this commits to the revenue sharing feature and there, create the PR for the server, since in the revenue sharing is where is used, isn't it?

Why don't we create the PR directly against the revenue sharing branch? The objective is to merge that branch into the master branch by the end of the sprint 🤞🏼

package.json
```diff
@@ -45,17 +46,18 @@
     "web3-utils": "1.2.6"
   },
   "devDependencies": {
-    "@openzeppelin/test-helpers": "0.5.10",
     "@openeth/truffle-typings": "0.0.6",
-    "@types/node": "^18.0.0",
-    "@typescript-eslint/eslint-plugin": "~4.28.3",
-    "@typescript-eslint/parser": "~4.28.3",
+    "@openzeppelin/test-helpers": "0.5.10",
```
jurajpiar, line 50: remove if not used

src/RelayPricer.ts
jurajpiar: add test for this.

src/RelayPricer.ts
```diff
@@ -0,0 +1,28 @@
...
+        return sourcePrice / targetPrice;
```
jurajpiar, line 26: I don't get this. The method is getPrice, the test says it should return the currency conversion price, but what it returns is an exchange rate. What am I missing?

src/api/CoinBase.ts
```diff
@@ -0,0 +1,26 @@
+import { CoinBaseResponse, SourceApi } from '../types/RelayPricer';
+import fetch, { Response } from 'node-fetch';
+
+export default class CoinBase implements SourceApi {
+    // we should get this address from config file or enviroments
+    private readonly URL: string = 'https://api.coinbase.com/v2/exchange-rates';
```
jurajpiar, line 6: just move it to a constant outside of the class for now, eg: `const COINBASE_URL  = ...`

src/api/CoinBase.ts
jurajpiar, line 17: you can use `response.ok`

jurajpiar: also I'd define this the other way around so that you have the return value as a green path, i.e, if not ok throw, otherwise continue

src/api/CoinBase.ts
```diff
@@ -0,0 +1,26 @@
...
+        const upperCaseOutput = output.toUpperCase();
+        const response: Response = await fetch(
+            `${this.URL}?currency=${input}`,
+            {
+                method: 'get'
+            }
+        );
+
+        if (response.status === 200) {
+            const result: CoinBaseResponse = await response.json();
```
jurajpiar, line 18:
```suggestion
            const {data: {rates}}: CoinBaseResponse = await response.json();
```

src/api/CoinBase.ts
```diff
@@ -0,0 +1,26 @@
...
+            if (!(upperCaseOutput in result.data.rates)) {
+                throw new Error('Output not available');
+            }
+            return Number(result.data.rates[upperCaseOutput]);
```
jurajpiar, line 22: just a knitpicking optimisation, but I think it would be better as:
```suggestion
            const rate = rates.find(rate => rate.toUpperCase() === output.toUpperCase()); // just to make sure the casing is the same, apply to both
            if (!rate) {
                throw new Error(`Exchange rate for currency pair ${input}/${output} is not available`);
            }
            return Number(rate);
```

src/api/CoinBase.ts
```diff
@@ -0,0 +1,26 @@
...
+        }
+        throw new Error('Input not available');
```
jurajpiar, line 24:
```suggestion
        throw new Error(`Coingecko API does not recognise given currency ${input}`);
```

src/api/CoinBase.ts
```diff
@@ -0,0 +1,26 @@
+import { CoinBaseResponse, SourceApi } from '../types/RelayPricer';
+import fetch, { Response } from 'node-fetch';
+
+export default class CoinBase implements SourceApi {
+    // we should get this address from config file or enviroments
+    private readonly URL: string = 'https://api.coinbase.com/v2/exchange-rates';
+
+    async query(input: string, output: string): Promise<number> {
```
jurajpiar, line 8:
```suggestion
    async query(sourceCurrency: string, targetCurrency: string): Promise<number> {
```

src/api/CoinBase.ts
jurajpiar: however, consider bubbling these errors up, so that you don't have to define the message for each source.

src/types/RelayPricer.ts
```diff
@@ -0,0 +1,10 @@
+export interface SourceApi {
+    query(input: string, ouput: string): Promise<number>;
```
jurajpiar, line 2:
```suggestion
    query(sourceCurrency: string, targetCurrency: string): Promise<number>;
```

jurajpiar, line 2: I would suggest using BigNumber for the return type as some ratios may be rather small (for example ratio of 1 iranian rial to btc is around 0.000000001100)

src/types/RelayPricer.ts
```diff
@@ -0,0 +1,10 @@
...
+}
+
+export type CoinBaseResponse = {
+    data: {
+        currency: string;
+        rates: Record<string, string>;
+    };
+};
```
jurajpiar, line 10: This should be inside the coinbase-specific file

src/RelayPricer.ts
franciscotobar: I updated it, I had a conversion with Andres regarding it and he mention something like that, but to keep simple I decided to not do the implementation, yet.

test/RelayPricer.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    describe('getPrice', async () => {
+        beforeEach(() => {
+            fakeSourceApi = stubInterface<SourceApi>();
+            pricer = new RelayPricer(fakeSourceApi);
+        });
+
+        it('should initiate sourceApi', async () => {
+            const coinBase = new CoinBase();
```
jurajpiar, line 38: I wouldn't use any concrete implementation here, but instead create a mock object from your `SourceApi` interface.

src/RelayPricer.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    async getPrice(
+        sourceCurrency: string,
+        targetCurrency: string,
+        intermediateCurrency?: string
+    ): Promise<number> {
+        const intermediary = intermediateCurrency
+            ? intermediateCurrency
+            : 'USD';
```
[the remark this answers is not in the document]
franciscotobar: updated

src/RelayPricer.ts
franciscotobar: On my understanding, the currency conversion price is the same as exchange rate, I could rename it.

test/RelayPricer.test.ts
```diff
@@ -0,0 +1,104 @@
...
+import { SourceApi } from '../src/types/RelayPricer';
+
+chai.use(chaiAsPromised);
+
+describe('RelayPricer', () => {
+    let fakeSourceApi: StubbedInstance<SourceApi>;
+    let pricer: RelayPricer;
+    const FAKE_PRICE_1 = 23345.834630269488;
+    const FAKE_PRICE_2 = 0.0775886924643371;
+    const FAKE_PRICE_CONVERSION_1 = 300892.2291221775;
```
jurajpiar, line 16: ah, I see what happened.
Perhaps, don't call it price conversion, I think that's misleading. Instead I'd call it exchange rate (or `xRate` for short) Also define what rate that is, eg `xRateUsdRbtc`.
also fake price name is confusing. This is a test. all data is fake, so no need to call it that.

src/api/CoinBase.ts
franciscotobar: When you say this, you mean like to constant class?

test/RelayPricer.test.ts
jurajpiar: also don't just make numbers of which calculations you're testing up. that can lead to false positives more easily.
if you want to test different values, have different made up currencies and calculate their ratio.

src/api/CoinBase.ts
franciscotobar: updated

src/api/CoinBase.ts
franciscotobar: updated

test/RelayPricer.test.ts
jurajpiar: one last thing here, I'd say that only constants defined outside of all scope (within a module, ie file) should be named with all caps. Meaning, that constants defined inside a describe block should follow normal lower camel casing.

src/api/CoinBase.ts
franciscotobar: rates it not an array. its an object. Ill reuse the error message.

src/api/CoinBase.ts
franciscotobar: updated, for simplicity I leave the errors like that, since so far there is only one source

src/api/CoinBase.ts
franciscotobar: updated

src/RelayPricer.ts
```diff
@@ -0,0 +1,28 @@
...
+
+        const [sourcePrice, targetPrice] = await Promise.all([
```
jurajpiar, line 19: by price you mean exchange rate, right?

src/RelayPricer.ts
jurajpiar: actually, I only just realised what this is doing, lol.
so here we're returning ratio of ratios? I think that's not going to work is it? Isn't it just same as saying for example:
$EUR/JPY == EUR/USD/USD/JPY$

src/RelayPricer.ts
franciscotobar: yes, I will rename the variables.

test/RelayPricer.test.ts
franciscotobar: updated

src/types/RelayPricer.ts
franciscotobar: What would be your recommendation for the naming, I have some doubts about it. Since the class is called RelayPricer and I named the type file as RelayPricer also, this could be confusing? I also have seen that some of the Class does not have a specific file and they declare the types inside the class file but I dont like this approach.

src/types/RelayPricer.ts
franciscotobar: I wanted to discuss before adding an extra bigNumber library but I did some test with bignumber.js and it works properly. I did not use BN since it does not work with decimals.

franciscotobar: updated

src/RelayPricer.ts
franciscotobar: updated

src/RelayPricer.ts
```diff
@@ -0,0 +1,28 @@
...
+            this.sourceApi.query(sourceCurrency, intermediary),
+            this.sourceApi.query(targetCurrency, intermediary)
+        ]);
+        if (sourcePrice === 0 || targetPrice === 0) {
+            throw new Error('price cannot be zero');
```
jurajpiar, line 24: the message is a bit meaningless for the caller of the method I'd say

src/api/CoinBase.ts
franciscotobar: updated

src/api/CoinBase.ts
franciscotobar: updated

src/types/RelayPricer.ts
franciscotobar: updated

test/RelayPricer.test.ts
franciscotobar: updated

src/RelayPricer.ts
franciscotobar: updated

package.json
franciscotobar: They are used inside of other classes

src/RelayPricer.ts
franciscotobar: updated

src/RelayPricer.ts
franciscotobar: renaming but the exchange works like the following:

sourceCurrency=EUR;
targetCurrency=JPY;
intermediateCurrency=USD;

sourceExchangeRate=USD/EUR
targetExchangeRate=USD/JPY

sourceExchangeRate/targetExchangeRate

src/types/RelayPricer.ts
franciscotobar: updated

src/RelayPricer.ts
franciscotobar: updated

src/RelayPricer.ts
jurajpiar: my bad 👍

src/RelayPricer.ts
jurajpiar: how about 
```
sourceApi `Currency conversion for pair ${sourceCurrency}:${targetCurrency} not found in current source api`);
```
or if we were to add a property to SourceApi interface: 
```
sourceApi `Currency conversion for pair ${sourceCurrency}:${targetCurrency} not found in ${this.sourceApi.name}`);
```
or further, if we were to make it accept a number of source apis: 
```
sourceApi `Currency conversion for pair ${sourceCurrency}:${targetCurrency} not found in any currently used apis. Use `listSourceApis` to list these.`);
```

which brings me to a new question: "SourceApi"... is that a meaningful name? Perhaps better would be "PriceApi" or "ExchangeApi" or "CurrencyConversionApi"? That would describe its job, whereas "source" is a bit vague.

package.json
jurajpiar: Screenshot 2022-08-26 at 2 27 03 pm
it sure doesn't look like it

src/RelayPricer.ts
jurajpiar: Conversation continues in https://github.com/rsksmart/rif-relay-client/pull/28#discussion_r956041187

test/RelayPricer.test.ts
```diff
@@ -35,69 +35,82 @@ describe('RelayPricer', () => {
...
+            expect(pricer).to.have.property('sourceApi', fakeSourceApi);
         });
 
-        it('should return source currency conversion price', async () => {
-            buildFakePrices(FAKE_PRICE_1, FAKE_PRICE_2);
-            const price = await pricer.getPrice(
-                SOURCE_CURRENCY,
-                TARGET_CURRENCY
+        it('should fail if sourceApi is null', async () => {
+            expect(() => new RelayPricer(null)).to.throw(
```
jurajpiar, line 44:
```suggestion
            expect(() => new RelayPricer(undefined as unknown as SourceApi)).to.throw(
```
I'm surprised TS didn't complain about it

src/RelayPricer.ts
```diff
@@ -1,28 +1,38 @@
...
+    async getExchangeRate(
         sourceCurrency: string,
         targetCurrency: string,
         intermediateCurrency?: string
-    ): Promise<number> {
+    ): Promise<BigNumber> {
         const intermediary = intermediateCurrency
             ? intermediateCurrency
-            : 'USD';
+            : INTERMEDIATE_CURRENCY;
```
jurajpiar, line 24:
```
const intermediary = intermediateCurrency ?? INTERMEDIATE_CURRENCY;
```
;)

jurajpiar (approved):
Looking good

src/RelayPricer.ts
ironFe93: I vote for ExchangeApi

package.json
franciscotobar: updated

PR #29, opened 2022-08-17
PP-127: no-collector

ironFe93:
## What

- removed collector fields
- relay worker removed from transaction details
- added methods to call ping handler
- collector contract and relay worker are now obtained by pinging the server

## Why

- There is no reason why anyone that uses the client should have to specify the collector contract, or even the relay worker since the rif-relay setup is confirmed to be single tenant.

## Refs

- PP-91, PP-127

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,8 +27,8 @@
     ]
   },
   "dependencies": {
-    "@rsksmart/rif-relay-common": "file:../rif-relay-common",
-    "@rsksmart/rif-relay-contracts": "file:../rif-relay-contracts",
+    "@rsksmart/rif-relay-common": "https://github.com/rsksmart/rif-relay-common#PP-127/no-collector-field",
+    "@rsksmart/rif-relay-contracts": "https://github.com/rsksmart/rif-relay-contracts#PP-94/optional-collector",
     "abi-decoder": "~2.3.0",
     "axios": "~0.21.1",
     "eth-sig-util": "2.5.2",
--- a/package.json
+++ b/package.json
@@ -27,8 +27,8 @@
     ]
   },
   "dependencies": {
-    "@rsksmart/rif-relay-common": "https://github.com/rsksmart/rif-relay-common#PP-127/no-collector-field",
-    "@rsksmart/rif-relay-contracts": "https://github.com/rsksmart/rif-relay-contracts#PP-94/optional-collector",
+    "@rsksmart/rif-relay-common": "https://github.com/rsksmart/rif-relay-common#PP-94/revenue-sharing-model",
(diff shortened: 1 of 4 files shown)


src/RelayClient.ts
```diff
@@ -809,14 +811,16 @@ export class RelayClient {
                 gasPrice,
                 callVerifier,
                 callForwarder: forwarderAddress,
-                relayWorker
+                feesReceiver: collectorAddress
```
franciscotobar, line 814: Would not be better to standardise the name to feeesReceiver?

ironFe93: Had exactly the same thought, but we might keep the distinction for explicitness. After all, the fees receiver is not always the collector contract.

antomor (changes requested):
@ironFe93 thank you for the changes. I've just left some minor comments.

src/RelayClient.ts
```diff
@@ -766,16 +768,16 @@ export class RelayClient {
         transactionDetails: EnvelopingTransactionDetails
     ): Promise<DeployTransactionRequest> {
         const forwarderAddress = this.resolveForwarder(transactionDetails);
-
         const senderNonce: string =
             await this.contractInteractor.getFactoryNonce(
                 forwarderAddress,
                 transactionDetails.from
             );
+        log.info({ senderNonce });
```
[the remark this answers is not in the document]
ironFe93: This should be removed, as I used it for testing purposes

src/RelayClient.ts
ironFe93: Ended up uses feesReceiver after all

src/Enveloping.ts
```diff
@@ -197,13 +200,12 @@ export default class Enveloping {
                 gas: gasToSend,
                 nonce: (await this.getSenderNonce(forwarder)).toString(),
                 tokenContract: tokenContract,
-                collectorContract: collectorContract ?? zeroAddr,
```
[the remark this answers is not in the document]
ironFe93: I didn't notice. Will remove.

package.json
```diff
@@ -27,8 +27,8 @@
     ]
   },
   "dependencies": {
-    "@rsksmart/rif-relay-common": "https://github.com/rsksmart/rif-relay-common#PP-94/revenue-sharing-model",
-    "@rsksmart/rif-relay-contracts": "https://github.com/rsksmart/rif-relay-contracts#PP-94/revenue-sharing-model",
+    "@rsksmart/rif-relay-common": "https://github.com/rsksmart/rif-relay-common#PP-127/no-collector-field",
+    "@rsksmart/rif-relay-contracts": "https://github.com/rsksmart/rif-relay-contracts#PP-94/optional-collector",
```
antomor, line 31: @ironFe93 We need to update these dependencies before merging the code.

antomor (approved):
LGTM

PR #31, opened 2022-08-30
PP-288/rif-rbtc

franciscotobar:
## What

- Integration of price feeder for RBTC-RIF

## Why

- Needed for revenue sharing validation process

## Refs
- [PR #288 (Implement a token-RBTC conversion)](https://rsklabs.atlassian.net/browse/PP-288)
- https://github.com/rsksmart/rif-relay-contracts/pull/53
- https://github.com/rsksmart/rif-relay-common/pull/31


src/api/CoinBase.ts
```diff
@@ -0,0 +1,63 @@
...
+    getApiTokenName(tokenSymbol: string): string {
+        if (!tokenSymbol) {
+            throw Error('ExchangeApi cannot map a token with a null value');
+        }
+        const resultMapping = CURRENCY_MAPPING[tokenSymbol];
+        if (resultMapping) {
+            return resultMapping;
+        }
+        return tokenSymbol;
+    }
```
AndresQuijano, line 30: If the tokenSymbol is not in the CURRENCY_MAPPING, this function returns the tokenSymbol itself. How can a caller distinguish between the case when the tokenSymbol is not in the CURRENCY_MAPPING and the case when the API name is the same as the tokenSymbol?

src/api/CoinBase.ts
```diff
@@ -0,0 +1,63 @@
...
+    RBTC: 'RBTC',
+    TKN: 'RIF'
+};
+
+export class CoinBase implements ExchangeApi {
+    getApiTokenName(tokenSymbol: string): string {
+        if (!tokenSymbol) {
+            throw Error('ExchangeApi cannot map a token with a null value');
+        }
+        const resultMapping = CURRENCY_MAPPING[tokenSymbol];
```
AndresQuijano, line 25: Have you tested this with lower/upper case tokenSymbol?

src/api/CoinBase.ts
```diff
@@ -0,0 +1,63 @@
...
+        if (resultMapping) {
+            return resultMapping;
+        }
+        return tokenSymbol;
+    }
+
+    async query(
```
AndresQuijano, line 32: Sorry man I know we have talked about this function a lot of times before but just now I think this is not the best name for it.

src/api/CoinBase.ts
```diff
@@ -0,0 +1,63 @@
...
+        const response: Response = await fetch(
+            `${URL}?currency=${sourceCurrency}`,
+            {
+                method: 'get'
+            }
+        );
+
+        if (!response.ok) {
+            throw Error(
+                `Coinbase API does not recognise given currency ${sourceCurrency}`
```
AndresQuijano, line 46: This error message is so specific. Are you sure this is the only reason that makes the response not OK?

antomor (commented):
Is this PR intended to replace #28?

src/api/CoinBase.ts
```diff
@@ -0,0 +1,63 @@
...
+    data: {
+        currency: string;
+        rates: Record<string, string>;
+    };
+};
+
+const CURRENCY_MAPPING: Record<string, string> = {
+    RIF: 'RIF',
+    RBTC: 'RBTC',
+    TKN: 'RIF'
```
antomor, line 17: I'd add a comment here to specify that this is for testing purposes only.
Or eventually, we could implement an ExchangeApi for the TestToken only in another task. WDYT?

franciscotobar: I think that the comment will be enough for now. 

The TestToken Exchange Api was my first approach but by implementing it, we would also need to change the way of how is implemented now since the ExchangeApi to be used, would be based on the token and therefore a selection process would be needed before.

src/api/CoinBase.ts
franciscotobar: Ill be updating it, you are right it could lead to wrong information

src/api/CoinBase.ts
franciscotobar: I will be updating with a different implementation, that could simplify it.

src/api/CoinBase.ts
franciscotobar: Ill be renaming it with a more descriptive name.

src/api/CoinBase.ts
franciscotobar: I don't think that the caller should care about this. If the caller needs his token to be in the mapping, he will need to add it manually.

franciscotobar: updated

src/api/CoinBase.ts
franciscotobar: updated

src/api/CoinBase.ts
franciscotobar: updated

src/api/CoinBase.ts
franciscotobar: updated

src/api/CoinBase.ts
franciscotobar: updated

franciscotobar:
Yes it is intended to replace that PR

AndresQuijano (approved):

src/api/TestExchange.ts
```diff
@@ -0,0 +1,37 @@
...
+        USD: '23345.834630269488'
+    }
+};
+
+const CURRENCY_MAPPING: Record<string, string> = {
+    TKN: 'TKN',
+    RBTC: 'RBTC'
+};
+
+export class TestExchange extends BaseExchangeApi implements ExchangeApi {
```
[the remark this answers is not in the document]
ironFe93: I guess this test class is meant to support the test token TKN. But I still don't think we should provide such a class. Instead, it should be up to whoever implements the Pricer to mock a response. Also if you are including this class, it should still be tested, but you probably didn't include a test because this is a test class by itself.

src/types/ExchangeApi.ts
```diff
@@ -0,0 +1,36 @@
...
+export interface ExchangeApi {
+    getApiTokenName(tokenSymbol: string): string;
+
+    queryExchangeRate(
+        sourceCurrency: string,
+        targetCurrency: string
+    ): Promise<BigNumber>;
+}
+
+export abstract class BaseExchangeApi {
```
ironFe93, line 12: I find it a bit confusing to put a class inside a /types folder. Would move it next to the other ExchangeApis.

ironFe93 (commented):
tldr: wouldn't include TestExchange.ts

src/api/CoinBase.ts
```diff
@@ -0,0 +1,67 @@
+import fetch, { Response } from 'node-fetch';
+import BigNumber from 'bignumber.js';
+import { BaseExchangeApi, ExchangeApi } from '../types/ExchangeApi';
+
+const URL = 'https://ap.coinbase.com/v2/exchange-rates';
```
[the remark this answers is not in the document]
franciscotobar: updated, it was a typo

src/types/ExchangeApi.ts
franciscotobar: updated

src/api/TestExchange.ts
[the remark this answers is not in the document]
franciscotobar: correct I did not test it because its a test class but I could copy the coinBase test's that could test this class. Im missing unit testing for the abstract class but it would also be the same test of coinBase. 

Since we are also implementing the pricer, this TestExchange was created to mock the price of the TKN that we use as testing. Or even we could not implement it and just map the TKN with RIF.

package.json
```diff
@@ -27,16 +27,18 @@
     ]
   },
   "dependencies": {
-    "@rsksmart/rif-relay-common": "https://github.com/rsksmart/rif-relay-common#PP-94/revenue-sharing-model",
-    "@rsksmart/rif-relay-contracts": "https://github.com/rsksmart/rif-relay-contracts#PP-94/revenue-sharing-model",
+    "@rsksmart/rif-relay-common": "https://github.com/rsksmart/rif-relay-common#PP-288/rif-rbtc",
+    "@rsksmart/rif-relay-contracts": "https://github.com/rsksmart/rif-relay-contracts#PP-288/rif-rbtc",
```
antomor, line 31: Dependencies need be updated before merging the PR

test/RelayPricer.test.ts
```diff
@@ -0,0 +1,146 @@
...
+        pricer = new RelayPricer();
+    });
+
+    describe('findAvailableApi', () => {
+        it('should return available api', () => {
+            const availableApi = pricer.findAvailableApi(sourceCurrency);
+            assert.containsAllKeys(availableApi, ['api', 'tokens']);
+        });
+
+        it('should fail is there is no available api', () => {
```
antomor, line 32:
```suggestion
        it('should fail if there is no available api', () => {
```

test/RelayPricer.test.ts
```diff
@@ -0,0 +1,146 @@
...
+            expect(() => pricer.findAvailableApi(targetCurrency)).to.throw(
+                `There is no available API for token ${targetCurrency}`
+            );
+        });
+
+        it('should fail token is null', () => {
```
antomor, line 38:
```suggestion
        it('should fail if token is null', () => {
```

test/RelayPricer.test.ts
```diff
@@ -0,0 +1,146 @@
...
+            expect(() => pricer.findAvailableApi(null)).to.throw(
+                `There is no available API for token ${null}`
+            );
+        });
+
+        it('should fail token is empty', () => {
```
antomor, line 44:
```suggestion
        it('should fail if token is empty', () => {
```

src/api/TestExchange.ts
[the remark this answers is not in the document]
antomor: @ironFe93 We ended up implementing this for two main reasons:
- to support our TestToken with the RelayPricer
- to start creating the structure to support later on different exchange APIs.

I know that a possible alternative could have been just having the value hard-coded without creating the additional class, but in this way we are following the same structure will be followed when a new exchange API will be implemented.

PR #36, opened 2022-10-26
PP-508/fees-from-server

franciscotobar:
## What

-   Consume the endpoint from the Relay Server for gas estimation

## Why

-  Simplify the gas estimation
- Centralize the source of the gas estimation

## Refs
-   [Jira Issue](https://rsklabs.atlassian.net/browse/PP-508)

### DIFF
--- a/src/HttpWrapper.ts
+++ b/src/HttpWrapper.ts
@@ -50,10 +50,7 @@ export default class HttpWrapper {
         }
     }
 
-    async sendPromise(
-        url: string,
-        jsonRequestData?: unknown
-    ): Promise<unknown> {
+    async sendPromise<T>(url: string, jsonRequestData?: unknown): Promise<T> {
         if (this.logreq) {
             console.log(
                 'sending request:',
@@ -62,7 +59,7 @@ export default class HttpWrapper {
             );
         }
 
-        const response = await this.provider.request({
+        const response = await this.provider.request<T>({
             url,
             method: jsonRequestData != null ? 'POST' : 'GET',
             data: jsonRequestData
--- a/src/HttpWrapper.ts
+++ b/src/HttpWrapper.ts
@@ -50,7 +50,10 @@ export default class HttpWrapper {
         }
     }
 
-    async sendPromise(url: string, jsonRequestData?: any): Promise<any> {
+    async sendPromise(
+        url: string,
+        jsonRequestData?: unknown
+    ): Promise<unknown> {
         if (this.logreq) {
             console.log(
                 'sending request:',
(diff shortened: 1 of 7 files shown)


src/RelayClient.ts
```diff
@@ -313,6 +324,8 @@ export class RelayClient {
                     relayWorkerAddress,
                     feesReceiver
                 )) + Number(trxDetails.tokenGas);
+
+            //FIXME check why two times its being multiply by ESTIMATED_GAS_CORRECTION_FACTOR
```
antomor, line 328: Is this still valid? do we plan to fix it in a different task?

test/RelayClient.test.ts
```diff
@@ -146,18 +153,84 @@ describe('RelayClient', () => {
...
+            restore();
+        });
+
+        it('should estimate relay gas limit', async function () {
+            const estimation = await relayClient.estimateGasLimit(
+                transactionDetails,
+                true
             );
+            expect(estimation).to.be.equal(responseRelayEstimation);
         });
```
antomor, line 234: I think we should add a test to cover different scenarios like:
- without signature
- deployment vs relay
- http client returns an error

antomor (changes requested):

src/RelayClient.ts
franciscotobar: fixed

test/RelayClient.test.ts
franciscotobar: Im including the tests, but I did not do it since this scenarios are tested in the gasEstimator and I thought that this type of scenarios will not be testing what this method actually do, that is basically build the request.

src/HttpWrapper.ts
```diff
@@ -50,7 +50,10 @@ export default class HttpWrapper {
         }
     }
 
-    async sendPromise(url: string, jsonRequestData?: any): Promise<any> {
+    async sendPromise(
+        url: string,
+        jsonRequestData?: unknown
+    ): Promise<unknown> {
```
antomor, line 56: I think this should be a generic function:
```suggestion
    async sendPromise<T>(
        url: string,
        jsonRequestData?: T
    ): Promise<T> {
        if (this.logreq) {
            console.log(
                'sending request:',
                url,
                JSON.stringify(jsonRequestData ?? {}).slice(0, LOGMAXLEN)
            );
        }

        const response = await this.provider.request<T>({
            url,
            method: jsonRequestData != null ? 'POST' : 'GET',
            data: jsonRequestData
        });
        return response.data;
    }
```

antomor (changes requested):

src/HttpWrapper.ts
franciscotobar: updated

antomor (approved):
LGTM

PR #34, opened 2022-10-04
feat(upgrade): change to ethers

jurajpiar:
## What

- replaces Web3 with ethers.js
- upgrades all other packages
- removes useless scripts
- strengthens linting and formatting
- adapts prod and test code

## Why

- demand for ethers.js
- technical dept

## Refs

- [Dependency update](https://rsklabs.atlassian.net/browse/PP-278)
  - [update rif-relay-client](https://rsklabs.atlassian.net/browse/PP-354)
    - [upgrade and change dependencies to ethers](https://rsklabs.atlassian.net/browse/PP-489)

PR #37, opened 2022-10-27
Feature/ethers/account manager

jurajpiar:
## What

- re-add AccountManager module
- adapt it to using ethers

## Why

- part of:

## Refs

- [ticket](https://rsklabs.atlassian.net/browse/PP-490)
- [parent](https://rsklabs.atlassian.net/browse/PP-489)


package.json
```diff
@@ -54,6 +54,7 @@
   },
   "dependencies": {
     "@metamask/eth-sig-util": "^5.0.0",
+    "@nomicfoundation/hardhat-toolbox": "^2.0.0",
```
jurajpiar, line 57: remove

package.json
```diff
@@ -45,7 +45,7 @@
     "prepack": "npmignore --auto",
     "prepare": "scripts/prepare",
     "prepublishOnly": "scripts/prepublishOnly",
-    "tdd": "npm run test -- --watch --watch-files src,test",
+    "tdd": "npm run test -- --watch --reporter mocha-clearscreen-reporter --watch-files src,test",
```
jurajpiar, line 48: remove

package.json
```diff
@@ -83,16 +84,16 @@
     "eslint": "~8.25.0",
     "eslint-config-prettier": "^8.5.0",
     "eslint-plugin-mocha": "^10.1.0",
-    "ethereum-waffle": "^3.4.4",
+    "eslint-plugin-prettier": "^4.2.1",
     "husky": "^8.0.1",
     "lint-staged": "^13.0.3",
-    "mocha": "^10.0.0",
+    "mocha": "^9.0.0",
+    "mocha-clearscreen-reporter": "^1.0.1",
```
jurajpiar, line 91: remove

src/AccountManager.ts
```diff
@@ -0,0 +1,134 @@
+import sigUtil, {
+  SignTypedDataVersion,
+  TypedMessage,
+} from '@metamask/eth-sig-util';
+import { providers, Signer, Wallet } from 'ethers';
+// import type { EnvelopingConfig } from './types/EnvelopingConfig';
```
jurajpiar, line 6: remove

src/AccountManager.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  private async _signWithProvider<T>(
+    data: TypedMessage<EnvelopingMessageTypes>
+  ): Promise<T> {
+    if (!(this._provider as providers.JsonRpcProvider).send) {
+      throw new Error(`Not an RPC provider`);
+    }
+
+    return (await (this._provider as providers.JsonRpcBatchProvider).send(
```
jurajpiar, line 118:
```suggestion
    return (await (this._provider as providers.JsonRpcBatchProvider).connect(this._signer).send(
```

jurajpiar: Not sure about this id which is just a date. Can anyone clarify whether it is used for anything?

Screenshot 2022-11-01 at 2 05 23 pm

franciscotobar: As far of my understanding. The Json RPC requests always needs an Id, which is meaningless for the blockchain but it could mean something to us. If they sent the date and its not use in other place, we can send anything.

src/typedRequestData.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+  verifier: string;
+  relayRequest:
+    | EnvelopingTypes.RelayRequestStruct
+    | EnvelopingTypes.DeployRequestStruct;
+  requestTypes: MessageTypeProperty[];
+};
+
+export const getEnvelopingRequestDataV4Field = ({
+  chainId,
+  verifier, // is this separated from the relay request by design?
```
jurajpiar, line 89: When would we want to use this with verifier external to the request that is being passed as third arg?

src/typedRequestData.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+  relayRequest,
+  requestTypes,
+}: GetRequestDataFieldProps): TypedMessage<EnvelopingMessageTypes> => ({
+  types: {
+    EIP712Domain: eip712DomainType,
+    RelayRequest: requestTypes,
+    RelayData: relayDataType,
+  },
+  primaryType: 'RelayRequest', // what if it's deploy request?
```
jurajpiar, line 98: Is this the correct primary type? Would have expected this to be `DeployRequest` for deploy requests...

test/types/AccountManager.test.ts
```diff
@@ -0,0 +1,312 @@
...
+use(sinonChai);
+use(chaiAsPromised);
+
+const sandbox = createSandbox();
+
+function createRandomAddress() {
+  const randomHex = crypto.randomBytes(20).toString('hex');
+
+  return getAddress(`0x${randomHex}`);
+}
```
jurajpiar, line 21: actually, this might be better using wallet.createWallet when I think about it...

test/types/AccountManager.test.ts
```diff
@@ -0,0 +1,312 @@
...
+        };
+
+        accountManager = new AccountManager(stubAdmin, stubProvider, 1);
+      });
+
+      describe('addAccount', function () {
+        it('should add wallet clone to local accounts', function () {
+          const wallet = Wallet.createRandom();
+
+          // sandbox.stub(providers.Provider, 'isProvider').callsFake(() => true)
```
jurajpiar, line 115: remove

src/typedRequestData.utils.ts
franciscotobar: The name was standardized but we could change it. If we are going to change it, we need to change in the contracts too, in the factories to be specific.

src/typedRequestData.utils.ts
franciscotobar: This verifier its not verifierContracts. This verifier is the address that we are trying to recover with the domain separator. It could be either the smartWallet address, or the smartWalletFactory.

src/AccountManager.ts
jurajpiar: just looked into the implementation of the jsonrpcprovider and it autoincrements it's own local id that begins with 42 lol.

package.json
franciscotobar: Im not sure what do you mean with by remove, well I do but not pretty sure.

src/typedRequestData.utils.ts
jurajpiar: I don't understand your answer, so I'll ask it differently:
the verifier is taken from the relay request and then the whole request is passed in as well in the account manager https://github.com/rsksmart/rif-relay-client/pull/34/files#diff-54caa818924335d964cb20a8998a3476391b16f046153c048a6147986d157e7fL92. Why is it important to keep those separate? Are there any scenarios where the verifier would not be taken from the request.

src/typedRequestData.utils.ts
jurajpiar: I mean, we have `RelayRequest` and `DeployRequest` types. Yet, here we hardcode it to the latter, and I just wonder: is that the expected primary type for all and if so, then why?

src/AccountManager.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+    this._accounts.push(wallet);
+  }
+
+  isDeployRequest({ request }: RelayOrDeployRequest): boolean {
+    return 'recoverer' in request;
+  }
+
+  async sign(relayRequest: RelayOrDeployRequest): Promise<string> {
+    const cloneRequest = { ...relayRequest }; // why does it have to be clone? don't seem to be mutated.
```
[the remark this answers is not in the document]
jurajpiar: ok I'm removing it.

src/typedRequestData.utils.ts
franciscotobar: The verifier is not taken from the relay request. This verifier is built in, in the domain separator. During the process of the signature recovering from the EIP712, the domain separator is needed to recover the external signature from the EOA from the client, not the worker.

src/typedRequestData.utils.ts
franciscotobar: Yes its expected to be the RelayRequest for the two types of transactions. Why Im not sure why they kept the same name, it could be changed, but that change includes a change in the contracts. 

This primary type that is declared here, serves to identify which is the struct has inside a nested struct. 

`RelayRequest(address relayHub,address from,address to,address tokenContract,address recoverer,uint256 value,uint256 nonce,uint256 tokenAmount,uint256 tokenGas,uint256 index,bytes data,RelayData relayData)RelayData(uint256 gasPrice,address feesReceiver,address callForwarder,address callVerifier)`

This line of code is inside the SmartWallet and SmartWalletFactory. As you can see the RelayRequest has inside of it the RelayData, and if we want we could change those names.

src/typedRequestData.utils.ts
jurajpiar: [see this line](https://github.com/rsksmart/rif-relay-client/blob/2d3f1668afa088d1e17e49ca7fbd515d191b93fa/src/AccountManager.ts#L92)
Screenshot 2022-11-01 at 4 21 46 pm
Screenshot 2022-11-01 at 4 23 24 pm

src/AccountManager.ts
```diff
@@ -0,0 +1,128 @@
...
+  addAccount(account: Wallet): void {
+    const wallet = new Wallet(account.privateKey, this._provider);
+    if (wallet.address !== account.address) {
+      throw new Error('invalid keypair');
+    }
+    this._accounts.push(wallet);
+  }
+
+  isDeployRequest({ request }: RelayOrDeployRequest): boolean {
+    return 'recoverer' in request;
```
franciscotobar, line 44: I would use the index instead of the recoverer.

src/typedRequestData.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+  verifyingContract: string,
+  chainId: number
+): EIP712Domain {
+  return {
+    name: 'string name,string version',
+    version: '2',
+    chainId: chainId,
+    verifyingContract: verifyingContract,
+  };
+}
```
franciscotobar, line 65:
```suggestion
function getDomainSeparator(
  verifyingContract: string,
  chainId: number
): EIP712Domain {
  return {
    name: domainSeparatorType.name,
    version: domainSeparatorType.version,
    chainId: chainId,
    verifyingContract: verifyingContract,
  };
}
```

franciscotobar: The domain separator structure includes a name and version, but you are putting the structure and it must match with the one inside the contracts:

https://github.com/rsksmart/rif-relay-contracts/blob/master/contracts/factory/SmartWalletFactory.sol#L84
https://github.com/rsksmart/rif-relay-contracts/blob/master/contracts/smartwallet/SmartWallet.sol#L27

test/AccountManager.test.ts
```diff
@@ -0,0 +1,292 @@
...
+            expectedMessage,
+            sandbox.match.any
+          );
+        });
+
+        it('should verify the request originator address against recovered typed signature', async function () {
+          sandbox
+            .stub(
+              accountManager as unknown as {
+                _recoverSignature: () => string;
```
franciscotobar, line 210: should we use this approach for stubbing private keys?

test/AccountManager.test.ts
```diff
@@ -0,0 +1,292 @@
...
+
+    describe('', function () {
+      let deployRequest: EnvelopingTypes.DeployRequestStruct;
+      let relayRequest: EnvelopingTypes.RelayRequestStruct;
+
+      let accountManager: AccountManager;
+      let stubProvider: providers.JsonRpcProvider;
+
+      beforeEach(function () {
+        const relayOrDeployRequest = {
```
franciscotobar, line 52: if we are stubbing the methods, why do we need the whole object? I try just to use the necessary properties and for ce the conversion to the required object.

src/typedRequestData.utils.ts
jurajpiar: First, my bad for not checking the code I copied over from https://github.com/rsksmart/rif-relay-contracts/blob/master/types/EIP712/TypedRequestData.ts#L82 (I would not likely have written
`{ ..., chainId: chainId, verifyingContract: verifyingContract }`).
Second, I really don't know what you're trying to convey here, sorry. Please rephrase.

test/AccountManager.test.ts
jurajpiar: Could you please clarify as to why you prefer to use only the necessary objects for given test?
I'd argue that:
1. it's better to build your objects that do not require change once (which, fair enough, is not what I am doing here, but probably should)
2. here specifically, it allowed me to split and see what is actually different between relay and deploy requests
3. by default you want to pass a complete object and if you want test for a lack of a property you remove it
4. reduces the amount of constants you need to then always build your object

src/AccountManager.ts
jurajpiar: 1. explain
2. the original implementation uses recoverer

test/AccountManager.test.ts
franciscotobar: There is no specific reason. Just trying to follow some standard. Ill change my test to use the whole object object.

test/AccountManager.test.ts
franciscotobar: no alternative os stubbing privates, on the unit tests that I have done, the private method was execute since I did not stub it.

src/typedRequestData.utils.ts
franciscotobar: In your code the domain separator is wrong. It should be like this:

```suggestion
export function getDomainSeparator(
  verifyingContract: string,
  chainId: number
): EIP712Domain {
  return {
    name: 'RSK Enveloping Transaction',
    version: '2',
    chainId: chainId,
    verifyingContract: verifyingContract,
  };
}
```

src/AccountManager.ts
franciscotobar: The recoverer function is not implemented yet and it could be optional since we are using always the zero address, but the index cannot be optional at all since its necessary.

src/typedRequestData.utils.ts
jurajpiar: that's just repeating yourself.
but actually I do see now what you mean. I don't know how the name parameter managed to get changed like that...

jurajpiar: done

test/AccountManager.test.ts
jurajpiar: ok then I don't know what the status of this comment is now....

test/AccountManager.test.ts
jurajpiar: ok so resolve this comment please if there is nothing to follow up on

jurajpiar: btw I moved it above the main describe block

src/AccountManager.ts
```diff
@@ -0,0 +1,128 @@
...
+    const data = getEnvelopingRequestDataV4Field({
+      chainId: this.chainId,
+      verifier: callForwarder,
+      relayRequest,
+      requestTypes: isDeploy ? deployRequestType : relayRequestType,
+    });
+
+    const wallet = this._accounts.find(
+      (account) => getAddress(account.address) === getAddress(from)
+    );
```
antomor, line 67:
```suggestion
    const fromAddress = getAddress(from)
    const wallet = this._accounts.find(
      (account) => getAddress(account.address) === fromAddress
    );
```

src/AccountManager.ts
```diff
@@ -0,0 +1,128 @@
...
+      } else {
+        signature = await this._signWithProvider(data, from);
+      }
+      rec = this._recoverSignature(data, signature);
+    } catch (error) {
+      throw new Error(
+        `Failed to sign relayed transaction for ${from}: ${error as string}`
+      );
+    }
+    if (getAddress(rec) !== getAddress(from)) {
```
antomor, line 83:
```suggestion
    if (getAddress(rec) !== fromAddress) {
```

test/AccountManager.test.ts
antomor: @jurajpiar I think that the point here is: if I stub the private methods, how am I going to test them? Should we create unit tests for private methods? I'd say no unless strictly required, so we should probably execute the private methods without stubbing them, shouldn't we?

antomor (changes requested):

test/AccountManager.test.ts
jurajpiar: not if the private method doesn't contribute to the test case

src/AccountManager.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  async sign(relayRequest: RelayOrDeployRequest): Promise<string> {
+    const {
+      relayData: { callForwarder },
+      request: { from },
+    } = relayRequest as typeof relayRequest & {
+      relayData: { callForwarder: string };
+      request: { from: string };
+    }; // this is here just to remove PromiseOrValue<string> type from the def string properties
```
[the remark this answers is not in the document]
jurajpiar: because then we'd always have to either cast or convert the variable to string. The type exists for their (typechain) ethers v5 target to avoid multiple awaits in nested calls. This is not our case however.
But yea, I guess this casting may cause issue if in future this type changes. I'll just convert then.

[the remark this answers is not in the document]
jurajpiar: done

src/AccountManager.ts
jurajpiar: done

src/AccountManager.ts
jurajpiar: done

test/AccountManager.test.ts
antomor: >Also another issue is that all this private method does is calling an dependency. Now, normally the dependency should be stubbed, but I was unable to stub sigUtil directly, so I wrapped it in a method. Should you find a better solution I will be more than happy to change it.

got it, if you couldn't stub the dependency, it makes sense to me

antomor (approved):
LG

PR #39, opened 2022-11-02
feat(ethers): adapt HttpWrapper module

jurajpiar:
## What

- re-add Enveloping module
- adapt it to using ethers
- write tests

## Refs

- [Ticket](https://rsklabs.atlassian.net/browse/PP-494)

### DIFF
--- a/src/api/common/HttpWrapper.ts
+++ b/src/api/common/HttpWrapper.ts
@@ -1,0 +1,94 @@
+import axios, {
+  AxiosError,
+  AxiosInstance,
+  AxiosRequestConfig,
+  AxiosResponse,
+} from 'axios';
+import log, { LogLevelDesc } from 'loglevel';
+
+const LOGMAXLEN = 120;
+const DEFAULT_TIMEOUT = 30000;
+
+const getCircularReplacer = () => {
+  const seen = new WeakSet();
+
+  return (_: unknown, value: unknown) => {
+    if (typeof value === 'object' && value !== null) {
+      if (seen.has(value)) {
+        return;
+      }
+      seen.add(value);
+    }
+
+    return value;
+  };
+};
+const stringify = (something: unknown): string =>
+  JSON.stringify(something, getCircularReplacer(), 3);
+
+const interceptors = {
+  logRequest: {
+    onResponse: (response: AxiosResponse) => {
+      log.info(
+        `Got a response: ${response.config.url as string}${stringify(
+          response.data
+        ).slice(0, LOGMAXLEN)}`
+      );
+
+      return response;
+    },
+    onError: (error: AxiosError) => {
+      const { response, message, config } = error;
+      if (error.request) {
+        log.error(`Request failed: ${stringify(error.request)}`);
+      }
+      if (response) {
+        log.error(`Response error: ${stringify(response)}`);
+      }
+      if (!error.request && !response) {
+        log.error('Error while setting up request', stringify(message));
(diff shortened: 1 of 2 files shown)


src/api/common/HttpWrapper.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    if (this._isLogEnabled) {
+      this._httpClient.interceptors.response.use(
+        interceptors.logRequest.onResponse,
+        interceptors.logRequest.onError
+      );
+    }
+  }
+
+  async sendPromise(url: string, jsonRequestData?: unknown): Promise<unknown> {
```
antomor, line 64: This could probably be a generic function 
```suggestion
  async sendPromise<T>(url: string, jsonRequestData?: unknown): Promise<T> {
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rsksmart/rif-relay-client",
-  "version": "1.0.0-alpha.0",
+  "version": "1.0.0-alpha.1",
```
antomor, line 3: I wouldn't change the package version here, since we're going to change it before releasing it.

antomor (changes requested):
@jurajpiar Thank you for the changes, just left some minor comments

src/api/common/HttpWrapper.ts
jurajpiar: fair. will change

package.json
jurajpiar: I don't know. I find it a convenient way to create a draft PR before making any changes, so that I can move a task from to-do to in-progress.

jurajpiar: but tbh, I don't understand what you mean since you increased the version in your suggestion.

src/api/common/HttpWrapper.ts
jurajpiar: done

package.json
jurajpiar: changed

src/api/common/HttpWrapper.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    const response = await this._httpClient.request({
+      url,
+      method: jsonRequestData ? 'POST' : 'GET',
+      data: jsonRequestData,
+    });
+
+    return response?.data as T;
+  }
```
antomor, line 80: @jurajpiar thank you for the changes, but I'd keep the requestData as `unknown` and the return data as typed with `T`. That would allow us to specify the expected type in axios.request<T>.
Furthermore, I'm not sure if we need to use `response?.data`; I think that response should be always defined. Shouldn't it?
```suggestion
  async sendPromise<T>(url: string, jsonRequestData?: unknown): Promise<T> {
    if (this._isLogEnabled) {
      console.log(
        'Sending request:',
        url,
        jsonRequestData && stringify(jsonRequestData).slice(0, LOGMAXLEN)
      );
    }

    const response = await this._httpClient.request<T>({
      url,
      method: jsonRequestData ? 'POST' : 'GET',
      data: jsonRequestData,
    });

    return response.data;
  }
```

jurajpiar: haha I've done it the other way around :D didn't even notice...

jurajpiar: done

antomor (approved):
@jurajpiar thank you for the changes 🙏🏽

PR #40, opened 2022-11-07
PP-500: rewrite relayed transaction validator

ironFe93:
## What

- ethers upgrade

## Why

- ethers upgrade

## Refs


src/RelayedTransactionValidator.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    if (!from) {
+      throw Error('Transaction has no signer');
+    }
+
+    const isDeploy = (request as DeployTransactionRequest).relayRequest.request
+      .recoverer
+      ? true
+      : false;
```
antomor, line 54: we should probably create a utility function for this, since we often check if a request is a deploy or no. (e.g.: a similar code can be found in the AccountManager also)

src/RelayedTransactionValidator.ts
```diff
@@ -0,0 +1,95 @@
...
+      ? this.contractInteractor.encodeDeployCallABI(
+          (request as DeployTransactionRequest).relayRequest,
+          request.metadata.signature
+        )
+      : this.contractInteractor.encodeRelayCallABI(
+          (request as RelayTransactionRequest).relayRequest,
+          request.metadata.signature
+        );
+
+    // TODO: the relayServer encoder returns zero-length buffer for nonce=0.`
```
antomor, line 66: I think we missed the part related to this comment. In the previous version, there was a specific conversion from zero-length buffer to number.
```
const receivedNonce =
                transaction.nonce.length === 0
                    ? 0
                    : transaction.nonce.readUIntBE(
                          0,
                          transaction.nonce.byteLength
                      );
```

antomor: We may add a test to mock such response from the server.

antomor (changes requested):
@ironFe93 Apart from the comments I left, i've noticed that the lint check fails. Could you please have a look at it?

src/RelayedTransactionValidator.ts
jurajpiar: btw @antomor the code could be cleaner as:
```
const { nonce } = transaction;
const receivedNonce = nonce.length && nonce.readUIntBE(0, nonce.byteLength);
```

test/RelayedTransactionValidator.test.ts
```diff
@@ -0,0 +1,272 @@
+import { expect, use } from 'chai';
+import sinonChai from 'sinon-chai'
+import { createSandbox, SinonStubbedInstance } from 'sinon';
+import { RelayedTransactionValidator } from '../src/RelayedTransactionValidator';
+import ContractInteractor from '@rsksmart/rif-relay-common/dist/src/ContractInteractor';
+import chaiAsPromised from 'chai-as-promised';
+import type { DeployTransactionRequest, EnvelopingConfig } from '@rsksmart/rif-relay-common/dist/src';
```
jurajpiar, line 7: move the whole configuration including the `EnvelopingConfig` typing to the client please.

scripts/dist
```diff
@@ -2,4 +2,4 @@
 
 rm -rf dist
 
-npx tsc --build --clean
+npx tsc
```
jurajpiar, line 5: why removing this?

tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "emitDecoratorMetadata": true,
     "extendedDiagnostics":true,
     "noErrorTruncation": true,
-    "sourceMap": true
+    "sourceMap": true,
```
jurajpiar, line 16: you might want to fix your editor's linter, because trailing commas are not supported by json format afaik

tsconfig.json
```diff
@@ -13,6 +13,7 @@
...
   },
+  "exclude": ["./test"]
```
jurajpiar, line 18: if you add an exclude here, you have to put node_modules there too. otherwise it will try to compile them, as this is an override for the default exclude.
ref: https://www.typescriptlang.org/tsconfig#exclude

tsconfig.prettier.json
```diff
@@ -0,0 +1,18 @@
...
+    "strictPropertyInitialization": true,
+    "exactOptionalPropertyTypes": false,
+    "alwaysStrict": true,
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
+    "extendedDiagnostics":true,
+    "noErrorTruncation": true,
+    "sourceMap": true,
+  },
+}
```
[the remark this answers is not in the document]
ironFe93: let's do it

src/RelayedTransactionValidator.ts
ironFe93: image
I might need some help with these. Currently, nonce is a number, which means it has no property length, byteLength or much less readUintBE. Maybe there's a more appropiate type than just Transaction?

tsconfig.json
ironFe93: No such thing is currently happening. Let's get together and check it out?

scripts/dist
ironFe93: the --build flag does not seem to do anything different than just running tsc. The --clean flag just deletes the dist folder after compiling, so it is the same as running rm -rf dist. Again, let’s meet to discuss this.

src/RelayedTransactionValidator.ts
ironFe93: this is done

src/RelayedTransactionValidator.ts
ironFe93: this doesn't apply anymore since we are using Ethers Transactions

scripts/dist
ironFe93: hope the misunderstanding is clear now, going with tsd --build --clean and the tsc --p ./tsconfig.build.json

tsconfig.json
ironFe93: restructured

tsconfig.json
ironFe93: uhh nothing seems to be broken

test/RelayedTransactionValidator.test.ts
```diff
@@ -0,0 +1,272 @@
...
+import { BigNumber, Transaction, Wallet } from 'ethers';
+
+use(sinonChai);
+use(chaiAsPromised);
+
+const sandbox = createSandbox();
+const createRandomAddress = () => Wallet.createRandom().address;
```
franciscotobar, line 14: We should have this in a utility for testing

antomor (changes requested):
@ironFe93 Thank you for the changes. The PR LGTM, but you may need to rebase the branch, since there are conflicts.

test/RelayedTransactionValidator.test.ts
ironFe93: Will include in the next PR for sure

ironFe93:
@antomor rebase is done!

antomor (approved):
LGTM

PR #41, opened 2022-11-09
PP-508(hotfix): replace node-fetch with axios

antomor:
## What

- replace node-fetch with axios

## Why

- to make the CoinBase component isomorphic


src/api/CoinBase.ts
```diff
@@ -27,33 +28,28 @@ const CURRENCY_MAPPING: Record<string, string> = {
...
-            const {
-                errors: [error]
-            }: CoinBaseErrorResponse = await response.json();
+        try {
+            response = await axios.get(`${URL}?currency=${sourceCurrency}`);
+        } catch (error: unknown) {
+            const { response } = error as AxiosError;
             throw Error(
-                `CoinBase API status ${response.status}/${response.statusText}/${error.message}`
+                `CoinBase API status ${response.status}/${response.statusText}`
```
jurajpiar, line 46: I think it is possible to error without response:
```suggestion
                `CoinBase API status ${response?.status}/${response?.statusText}`
```
See: https://github.com/rsksmart/rif-relay-client/blob/45337c5a7c3958dcee5610c77a0054ae43606b11/src/api/common/HttpWrapper.ts#L40

Ref: https://axios-http.com/docs/handling_errors

src/api/CoinBase.ts
```diff
@@ -1,4 +1,5 @@
-import fetch, { Response } from 'node-fetch';
+import axios, { AxiosError } from 'axios';
+import { AxiosResponse } from 'axios';
```
jurajpiar, line 2: Merge

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rsksmart/rif-relay-client",
-  "version": "1.0.0-alpha.1",
+  "version": "1.0.0-alpha.2",
```
jurajpiar, line 3: If it is a hot fix should it not be 
```suggestion
  "version": "1.0.1-alpha.1",
```
?

src/api/CoinBase.ts
```diff
@@ -27,33 +28,28 @@ const CURRENCY_MAPPING: Record<string, string> = {
...
             throw Error(
-                `CoinBase API status ${response.status}/${response.statusText}/${error.message}`
+                `CoinBase API status ${response.status}/${response.statusText}`
             );
         }
 
         const {
             data: { rates }
-        }: CoinBaseResponse = await response.json();
+        }: CoinBaseResponse = await response.data;
```
jurajpiar, line 52: Hm so the object goes: response.data.data?

Also remove await.

jurajpiar (approved):
LG just look at the error handling.

package.json
antomor: no "stable" version has been released, so I think we should still move forward the next pre-release number instead of the minor one

src/api/CoinBase.ts
antomor: Done. 
However, once we migrate this to support ethers, we may need to unify the error handling to avoid code repetition.

src/api/CoinBase.ts
antomor: Done

package.json
jurajpiar: ok

src/api/CoinBase.ts
```diff
@@ -25,35 +26,77 @@ const CURRENCY_MAPPING: Record<string, string> = {
...
+            const { response, request } = error as AxiosError;
+            if (response) {
+                throw Error(
+                    `CoinBase API status ${response.status}/${response.statusText}`
+                );
+            }
+            if (request) {
+                throw Error('No response received from CoinBase API');
+            }
+            throw new Error('The request was not sent to the CoinBase API');
```
jurajpiar, line 94: why not just let the interceptor handle it?
 In any case, is it your intention to hide the original error? Just because this catch block does not seem to expose it anywhere.

jurajpiar: not an issue

PR #42, opened 2022-11-10
PP-495: Change the KnownRelaysManager to support ethers.js

ironFe93:
## What

- rewrite known relays manager

## Why

- code migration

## Refs

### DIFF
--- a/src/KnownRelaysManager.ts
+++ b/src/KnownRelaysManager.ts
@@ -50,8 +50,6 @@ export default class KnownRelaysManager {
 
   private readonly scoreCalculator: AsyncScoreCalculator;
 
-  private latestScannedBlock = 0;
-
   private relayFailures = new Map<string, RelayFailureInfo[]>();
 
   public relayLookupWindowParts: number;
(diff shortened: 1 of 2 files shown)


franciscotobar (approved):
LGTM.

src/KnownRelaysManager.ts
```diff
@@ -0,0 +1,292 @@
...
+        fromBlock,
+        toBlock,
+      });
+    });
+
+    return Promise.all(pastEventsPromises).then(
+      (pastEventArray: TypedEvent[][][]) => {
+        return pastEventArray.flat();
+      }
+    );
```
antomor, line 161: I don't think we're covering the exact same functionality we had before. By reading the previous code, it looks like there was a sort of retrial mechanism that we're missing now. I know that we aren't going to use this component, but I'd keep the same logic, just not to forget about it.

src/KnownRelaysManager.ts
```diff
@@ -0,0 +1,292 @@
...
+
+      const {
+        args: { relayManager },
+      } = genericEvent;
+
+      if (!genericEvent) {
+        log.info('No relayManager found in event');
+
+        return;
+      }
```
antomor, line 196: I think this may never happen, since we're doing the exact check at line 182. Am I missing something?

src/KnownRelaysManager.ts
```diff
@@ -0,0 +1,292 @@
...
+      relayUrl,
+    };
+    if (!relayFailures) {
+      this.relayFailures.set(relayUrl, [newFailureInfo]);
+    } else {
+      relayFailures.push(newFailureInfo);
+    }
+  }
+
+  getLastScannedBlock(): number {
```
antomor, line 289: It looks like this method isn't used anywhere

antomor (changes requested):
@ironFe93 thank you for the changes, I've left some doubts about the new code.

src/KnownRelaysManager.ts
ironFe93: Restored

src/KnownRelaysManager.ts
ironFe93: You are right, it was an oversight from my part. Removed.

src/KnownRelaysManager.ts
ironFe93: removed this along with the private member latestScannedBlock, which strangely isn’t used anywhere.

antomor (approved):
LGTM

PR #43, opened 2022-11-10
PP-570/pricer-case-sensitive

franciscotobar:
## What

- The relayPricer and API needs to handle case sensitive Currencies and Tokens. 
- Typing to guarantee the appropriate name.

## Why

- There are different type of casing for tokens and currencies and we need to be able to let the programmer know that the name that is using is valid.

## Refs

- (PP-570)[https://rsklabs.atlassian.net/browse/PP-570]

### DIFF
--- a/src/api/ExchangeApi.ts
+++ b/src/api/ExchangeApi.ts
@@ -1,11 +1,21 @@
 import BigNumber from 'bignumber.js';
 
+const avialableCurrencies = ['TRIF', 'RIF', 'RDOC', 'RBTC', 'TKN'] as const;
+
+export type FromCurrency = typeof avialableCurrencies[number];
+
+export type CurrencyMapping = Partial<Record<FromCurrency, string>>;
+
 export default abstract class BaseExchangeApi {
+    public readonly tokens: string[];
+
     constructor(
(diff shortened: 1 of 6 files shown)


src/api/ExchangeApi.ts
```diff
@@ -1,26 +1,39 @@
...
                 `${this.api} API cannot map a token with a null/empty value`
             );
         }
+
         const upperCaseTokenSymbol = tokenSymbol.toUpperCase();
-        const resultMapping = this.currencyMapping[upperCaseTokenSymbol];
-        if (!resultMapping) {
+
+        const currency =
+            this.currencyMapping[upperCaseTokenSymbol as FromCurrency];
```
ironFe93, line 28: would try to avoid the 'as FromCurrency' part but I'm probably being too nitpicky

ironFe93 (approved):
lgtm

franciscotobar:
you are right, I tried to avoid it several times but I need an opinion about it, maybe we can discuss it but if you see the history, I did a commit without forcing the typing but I change it.

antomor (approved):
LGTM

PR #45, opened 2022-11-11
feat(ethers): adapts HttpClient module

jurajpiar:
## What

- re-adds HttpClient module
- adapts and refactors it to using ethers
- adds tests

## Refs

- [Ticket](https://rsklabs.atlassian.net/browse/PP-493)


test/api/common/HttpClient.test.ts
```diff
@@ -0,0 +1,87 @@
+import type { EnvelopingTypes } from '@rsksmart/rif-relay-contracts/dist/typechain-types/contracts/RelayHub';
+import { expect, use } from 'chai';
+import { createSandbox } from 'sinon';
+import sinonChai from 'sinon-chai';
+import { HttpClient, HttpWrapper } from '../../../src/api/common';
```
jurajpiar, line 5: should really sort out absolute paths at some point to avoid double-dot hell

jurajpiar:
just realised that there is another method in the newer version of this file 🤦‍♂️

src/api/common/HttpClient.ts
```diff
@@ -0,0 +1,73 @@
...
+import type {
+    HubInfo,
+    RelayEstimation,
+    RelayOrDeployRequest
+} from 'src/common/relay.types';
+import type HttpWrapper from './HttpWrapper';
+
+const PATHS = {
+  GET_INFO: '/getaddr',
+  POST_REALY_REQEUST: '/relay',
```
antomor, line 11:
```suggestion
  POST_REALY_REQUEST: '/relay',
```

src/api/common/HttpClient.ts
```diff
@@ -0,0 +1,73 @@
...
+const VERIFIER_SUFFIX = '?verifier=';
+
+export default class HttpClient {
+  private readonly _httpWrapper: HttpWrapper;
+
+  constructor(httpWrapper: HttpWrapper) {
+    this._httpWrapper = httpWrapper;
+  }
+
+  async getHubInfo(relayUrl: string, verifier = ''): Promise<HubInfo> {
```
antomor, line 23: I'm afraid that the name isn't really appropriate, because we're not retrieve the hub info only.
What about `getChainInfo`?

src/api/common/HttpClient.ts
```diff
@@ -0,0 +1,73 @@
...
+    const verifierSuffix = verifier && `${VERIFIER_SUFFIX}${verifier}`;
+    const hubInfo: HubInfo = await this._httpWrapper.sendPromise(
+      relayUrl + PATHS.GET_INFO + verifierSuffix
+    );
+    if (!hubInfo) {
```
antomor, line 28: Why don't we check for the error here?

src/api/common/HttpClient.ts
```diff
@@ -0,0 +1,73 @@
...
+  ): Promise<string> {
+    // TODO: where does the response type come from?
+    const { signedTx, error } = await this._httpWrapper.sendPromise<{
+      signedTx: string;
+      error: string;
+    }>(relayUrl + PATHS.POST_REALY_REQEUST, request);
+    log.info('relayTransaction response:', signedTx, error);
+    if (error) {
+      throw new Error(`Got error response from relay: ${error}`);
+    }
```
antomor, line 48: Probably we could create a method to handle the error, since we're repeating the same code here and in the `estimateMaxPossibleGas`

antomor (changes requested):
@jurajpiar Thank you for the changes. Could you please review my comments? Apart from a typo, the others are related to the way we manage the error. We may probably avoid the code repetition there and create a single function for this.

src/api/common/HttpClient.ts
jurajpiar: Bit worried about [this error handling](https://github.com/rsksmart/rif-relay-server/blob/master/src/HttpServer.ts#L181) in the server, which also points out the uselessness of this check here on L28. We should probably aim at changing that to something more sensible.

src/api/common/HttpClient.ts
jurajpiar: actually, I reckon this should be an interceptor.

src/api/common/HttpClient.ts
jurajpiar: done

test/api/common/HttpClient.test.ts
antomor: shouldn't we configure `paths` in `tsconfig.json`?

jurajpiar: Let's handle that in a separate task tho

test/api/common/HttpClient.test.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      await expect(httpClient.getChainInfo(fakeURL)).to.be.rejectedWith(
+        'Got error response from relay: bar',
+      );
+    });
+  });
+
+  describe('relayTransaction', function () {
+    it(`should call httpWrapper.sendPromise with given url + ${RELAY_PATH}`, async function () {
+      const request = { foo: 'bar' } as unknown as EnvelopingTypes.DeployRequestStruct;
```
franciscotobar, line 85: Should not be using the whole DeployRequestStruct, to make the unit test more easy to understand?

antomor: IMHO not necessary, including additional properties may make me understand they're required for this test.

PR #46, opened 2022-11-11
PP-501: Adapts SmartWalletDiscovery module

jurajpiar:
## What

- re-adds SmartWalletDiscovery module
- adapts and refactors it to using ethers
- changes class to functions
- adds tests
- adds recommended jest-formatting rules - useful rules requiring padding around test-related blocks. could write our own in future using [padding-line-between-statements](https://eslint.org/docs/latest/rules/padding-line-between-statements) rule

## Refs

- [Ticket](https://rsklabs.atlassian.net/browse/PP-501)


src/smartWalletDiscovery.ts
```diff
@@ -0,0 +1,321 @@
...
+    ['bytes1', 'address', 'bytes32', 'bytes32'],
+    ['0xff', config.factory, salt, bytecodeHash]
+  );
+
+  const address = `0x${data.slice(26, data.length)}`;
+
+  return getAddress(address);
+};
+
+const hasAccActivity = async (
```
antomor, line 127: I'd rename it to `hasAccountActivity` to be clearer

src/smartWalletDiscovery.ts
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    swIndex++;
+  }
+
+  return swAccounts;
+}
+
+const discoverAccountsFromMnemonic = async (
+  configOveride: Config,
```
antomor, line 246: `configOverride`

src/smartWalletDiscovery.ts
```diff
@@ -0,0 +1,321 @@
...
+        nextHardened = true;
+        inactivityCountdown = config.eoaGap;
+
+        accounts = [
+          ...accounts,
+          {
+            eoaAccount,
+            swAccounts,
+          },
+        ];
```
antomor, line 298: It's not clear to me if we're covering [this case](https://github.com/rsksmart/rif-relay-client/blob/master/src/SmartWalletDiscovery.ts#L223)

src/smartWalletDiscovery.ts
jurajpiar: I've actually renamed it to represent an action: `checkHasActivity`

src/smartWalletDiscovery.ts
jurajpiar: yes, that's only there becuase they initialise the `discoveredAccount` object and then mutate it, whereas we use it directly.

src/smartWalletDiscovery.ts
jurajpiar: done

src/discovery/discovery.utils.ts
```diff
@@ -0,0 +1,314 @@
...
+  walletIndex: number,
+  bytecodeHash: string
+) => {
+  const isCustom = config.logic && config.logicParamsHash;
+  const salt = isCustom
+    ? solidityKeccak256(
+        ['address', 'address', 'address', 'bytes32', 'uint256'],
+        [
+          ownerEOA,
+          config.recoverer,
```
franciscotobar, line 110: Why is this information coming from the configuration? This should be a param not a configuration.

src/discovery/discovery.utils.ts
```diff
@@ -0,0 +1,314 @@
...
+import { langKo as ko } from '@ethersproject/wordlists/lib/lang-ko';
+import {
+  langZhCn as zh_cn,
+  langZhTw as zh_tw,
+} from '@ethersproject/wordlists/lib/lang-zh';
+import type { SmartWalletFactory } from '@rsksmart/rif-relay-contracts/dist/typechain-types/contracts/factory/SmartWalletFactory';
+import { ERC20__factory } from '@rsksmart/rif-relay-contracts/dist/typechain-types/factories/@openzeppelin/contracts/token/ERC20/ERC20__factory';
+import { SmartWalletFactory__factory } from '@rsksmart/rif-relay-contracts/dist/typechain-types/factories/contracts/factory';
+const SHA3_NULL_S =
+  '0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470'; //TODO: verify this is the correct way to get the hash. Copied from rif-relay-common (https://github.com/rsksmart/rif-relay-common/blob/f8e4ce8d90a979f3eec55a0eb5dbd5cc742c40e3/src/Constants.ts#L18-L19)
```
franciscotobar, line 24: If you apply a keccak-256 to an empty string, this is the value that it returns. Im not sure if need to have in a constant or generate dynamically.

src/discovery/discovery.utils.ts
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    const swAccounts = await findSWActivity(
+      eoaAccount,
+      config,
+      bytecodeHash,
+      provider
+    );
+
+    if (swAccounts.length > 0 || eoaActivityFound) {
```
franciscotobar, line 243: why are you validating again the eoaActivityFound?

src/discovery/discovery.utils.ts
```diff
@@ -0,0 +1,315 @@
...
+    const { address: eoaAccount } = rootNode.derivePath(
+      constructDerivationPath(eoaIndex, hardenedAccIdx)
+    );
+    const eoaActivityFound = await checkHasActivity(
+      eoaAccount,
+      config,
+      provider
+    );
+
+    if (eoaActivityFound) {
```
franciscotobar, line 232: why are you validating eoaActivityFound here?

src/discovery/discovery.utils.ts
jurajpiar: aaaaaah right... thanks!

src/discovery/discovery.utils.ts
jurajpiar: please expand your question. as it is I can only answer: why not?

franciscotobar: In line 243 you are validating the eoaActivityFound flag and restoring the gap. I don't think that the flag would change inside `findSWActivity`, therefore why is necessary to do this validating twice?

jurajpiar: ah well spotted tx

jurajpiar: done

src/discovery/discovery.utils.ts
jurajpiar: done

.eslintrc.cjs
```diff
@@ -2,6 +2,7 @@ module.exports = {
   extends: [
     "eslint:recommended",
     "plugin:mocha/recommended",
+    "plugin:jest-formatting/recommended",
```
antomor, line 5: why do we need these additinal rules?

src/smartWalletDiscovery.ts
antomor: You're right, my fault. Thank you for the clarification 👍🏼

antomor (approved):
@jurajpiar thank you for these changes, although we couldn't run it with the integration test, LGTM.
I know it's a matter of preferences, but I'm not a fan of `scr/discovery/discovery.ts` and `scr/discovery/discovery.utils.ts`. I'd avoid the repetition, especially in the utils file, but it's up-to-you 👍🏼

.eslintrc.cjs
jurajpiar: It was the fastest way to get the rules that ensure proper spacing between test blocks.
I could create custom padding-between-statements rules to capture this, but it would take extra effort.
See their implementation: https://github.com/dangreenisrael/eslint-plugin-jest-formatting/blob/master/src/rules/padding.ts

antomor: Fair enough, would you mind adding a comment to explain it?

jurajpiar:
> @jurajpiar thank you for these changes, although we couldn't run it with the integration test, LGTM. I know it's a matter of preferences, but I'm not a fan of `scr/discovery/discovery.ts` and `scr/discovery/discovery.utils.ts`. I'd avoid the repetition, especially in the utils file, but it's up-to-you 👍🏼

Agreed to change to:
```
|__discovery
     |__ index.ts  -> to expose the functions
     |__ utils.ts  -> to define all the functions 
```

antomor (changes requested):
@Juraj I think we can delete the file `test/utils.ts` since we're not using any more the `callFakeOnce` method.

.eslintrc.cjs
jurajpiar: done

antomor (approved):

PR #55, opened 2022-12-06
PP-605/RelayClient: _getInternalCallCost

franciscotobar:
## What

- Add relay client
- Re implement _getInternalCallCost

## Refs

- [PP-605](https://rsklabs.atlassian.net/browse/PP-605)


src/RelayClient.ts
```diff
@@ -0,0 +1,56 @@
...
+        return estimation;
+    }
+
+    private _applyGasCorrectionFactor(estimation: BigNumberish): BigNumber {
+        let bigEstimation = BigNumberJs(estimation.toString());
+        const bigGasCorrection = BigNumberJs(
+            ESTIMATED_GAS_CORRECTION_FACTOR
+        );
+
+        if (!bigGasCorrection.isEqualTo(1)) {
```
ironFe93, line 48: What's the harm in letting it apply a correction factor of 1?

jurajpiar:
rebased on latest `feature/ethers/dependencies`

src/RelayClient.ts
jurajpiar: also this check is unnecessary unless ESTIMATED_GAS_CORRECTION_FACTOR is a configuration property.

src/constants/relay.const.ts
```diff
@@ -0,0 +1 @@
+export const ESTIMATED_GAS_CORRECTION_FACTOR = '1';
```
jurajpiar, line 1: really shouldn't be a string if it is a number/BN.

jurajpiar (approved):
LG

src/constants/relay.const.ts
franciscotobar: its a bigNumber, I had it like that but Im not sure now. Why cannot be a string?

src/RelayClient.ts
franciscotobar: ESTIMATED_GAS_CORRECTION_FACTOR, its a configuration property.

src/RelayClient.ts
```diff
@@ -103,6 +105,45 @@ class RelayClient extends EnvelopingEventEmitter {
...
+    if (addCorrection) {
+      estimation = this._applyGasCorrectionFactor(estimation);
+    }
+
+    return estimation;
+  };
+
+  private _applyGasCorrectionFactor(estimation: BigNumberish): BigNumber {
+    let bigEstimation = BigNumberJs(estimation.toString());
+    const bigGasCorrection = BigNumberJs(ESTIMATED_GAS_CORRECTION_FACTOR);
```
antomor, line 139: I don't like the fact that here we're using this constant without having the possibility to configure it.
I think it would more flexible if we had an optional parameter `esimatedGasCorrectFactor` that if not specified uses the constant.
We should probably specify that parameter in the method `estimateInternalCallGas` so that the `_applyGasCorrectionFactor` can receive it from there.

src/constants/relay.const.ts
antomor: @franciscotobar I think Juraj was proposing to specify the type there as BigNumber directly. @jurajpiar was that what you meant?

antomor (changes requested):
@franciscotobar Thank you for these changes. My only doubt is related to the usage of constants, that cannot be changed elsewhere. For these factors, I'd prefer to have the possibility to change them, even if we don't specify it by default.

test/RelayClient.test.ts
```diff
@@ -40,11 +40,123 @@ const FAKE_TX_COUNT = 456;
...
+          gasPrice: '60000000',
+        },
+      } as EnvelopingTypes.RelayRequestStruct;
+      relayClient = new RelayClient();
+      provider = sandbox.createStubInstance(providers.JsonRpcProvider);
+      (relayClient as unknown as { _provider: providers.Provider })._provider =
+        provider;
+    });
+
+    afterEach(function () {
```
AndresQuijano, line 88: Is this the same afterEach on line 45?

test/RelayClient.test.ts
```diff
@@ -40,11 +40,123 @@ const FAKE_TX_COUNT = 456;
...
 
+const createRandomAddress = () => Wallet.createRandom().address;
+
 describe('RelayClient', function () {
   afterEach(function () {
     sandbox.restore();
   });
 
+  describe('constructor', function () {
+    it('should store provider', function () {
```
AndresQuijano, line 51: I don't get the idea of this test. You are not even calling the constructor. I'll ask you about it because perhaps I can learn one thing or two.

AndresQuijano, line 51: I don't get the idea of this test.  I would prefer check if getDefaultProvider() and getEnvelopingConfig() were called instead of setting the provider manually. I'll be asking you about this because perhaps I'm missing something that I can learn.

test/RelayClient.test.ts
franciscotobar: not really, it has different purposes

src/RelayClient.ts
franciscotobar: updated

src/RelayClient.ts
```diff
@@ -134,6 +139,48 @@ class RelayClient extends EnvelopingEventEmitter {
...
+    });
+
+    // The INTERNAL_TRANSACTION_ESTIMATE_CORRECTION is substracted because the estimation is done using web3.eth.estimateGas which
+    // estimates the call as if it where an external call, and in our case it will be called internally (it's not the same cost).
+    // Because of this, the estimated maxPossibleGas in the server (which estimates the whole transaction) might not be enough to successfully pass
+    // the following verification made in the SmartWallet:
+    // require(gasleft() > req.gas, "Not enough gas left"). This is done right before calling the destination internally
+    if (estimation.gt(internalEstimationCorrection)) {
+      estimation = estimation.sub(internalEstimationCorrection);
+    }
```
antomor, line 162: I think this part is shared with other [parts of the code](https://github.com/rsksmart/rif-relay-client/pull/60/files#diff-61c300ca57d8839a4c24a5d6700547d76cd72bca4c9ff8941705f5a7d4a0e983R207), so I'd include that logic (with the comment in an external function).

src/RelayClient.ts
```diff
@@ -134,6 +139,48 @@ class RelayClient extends EnvelopingEventEmitter {
...
+
+    if (addCorrection) {
+      estimation = this._applyGasCorrectionFactor(estimation);
+    }
+
+    return estimation;
+  };
+
+  private _applyGasCorrectionFactor(
```
antomor, line 171: If we move this function in utils, it may be used elsewhere

antomor (commented):
We could decide to extract those utility functions in this PR or in the other one.

AndresQuijano (approved):

PR #59, opened 2022-12-12
feat(ethers): unifies relayRequest composition in relay client

jurajpiar:
## What

- adds _getRelayRequestDetails function

## Why

- prevents building the request multiple times across the RelayClient class

## Refs

- [PP-617](https://rsklabs.atlassian.net/browse/PP-617)

### DIFF
--- a/src/RelayClient.ts
+++ b/src/RelayClient.ts
@@ -1,19 +1,31 @@
-import {
-  EnvelopingTypes,
-  IForwarder__factory,
-} from '@rsksmart/rif-relay-contracts';
+import { IForwarder__factory } from '@rsksmart/rif-relay-contracts';
+import type { PromiseOrValue } from '@rsksmart/rif-relay-contracts/dist/typechain-types/common';
 import { BigNumber as BigNumberJs } from 'bignumber.js';
-import { BigNumber, BigNumberish, getDefaultProvider, providers, utils } from 'ethers';
+import {
+  BigNumber,
+  BigNumberish,
+  constants,
+  getDefaultProvider,
+  providers,
+} from 'ethers';
+import { isAddress, solidityKeccak256 } from 'ethers/lib/utils';
 import AccountManager from './AccountManager';
 import type { EnvelopingConfig } from './common/config.types';
 import type { EnvelopingMetadata, HubInfo } from './common/relayHub.types';
 import type {
+  CommonEnvelopingRequestBody,
+  DeployRequestBody,
+  EnvelopingRequest,
   EnvelopingRequestData,
   RelayRequest,
   RelayRequestBody,
-  UserDefinedRelayRequest,
+  UserDefinedDeployRequestBody,
+  UserDefinedEnvelopingRequest,
+  UserDefinedRelayRequestBody,
 } from './common/relayRequest.types';
+import { isDeployRequest } from './common/relayRequest.utils';
 import type { EnvelopingTxRequest } from './common/relayTransaction.types';
+import { MISSING_CALL_FORWARDER } from './constants/errorMessages';
 import {
   ESTIMATED_GAS_CORRECTION_FACTOR,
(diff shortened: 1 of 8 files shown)


src/RelayClient.ts
```diff
@@ -44,16 +76,206 @@ class RelayClient extends EnvelopingEventEmitter {
...
+      envelopingRequest.relayData.callVerifier ||
+      this._envelopingConfig[
+        isDeployment ? 'deployVerifierAddress' : 'relayVerifierAddress'
+      ];
+
+    if (!callVerifier) {
+      throw new Error('No call verifier present. Check your configuration.');
+    }
+
+    const relayServerUrl = this._envelopingConfig.preferredRelays.at(0)?.url;
```
franciscotobar, line 126: During the relayTransaction, we ping/query the servers and try to execute the transaction using their services, Im guessing that the feesReceiver can be retrieve during that process, instead of get from the first server available.

src/RelayClient.ts
```diff
@@ -44,16 +76,206 @@ class RelayClient extends EnvelopingEventEmitter {
...
+    }
+
+    const nonce =
+      envelopingRequest.request.nonce ||
+      (await IForwarder__factory.connect(
+        callForwarder.toString(),
+        this._provider
+      ).nonce());
+
+    if (!nonce || BigNumber.from(nonce).isZero()) {
```
franciscotobar, line 176: nonce can be zero if its the first transaction, so the nonZero validation can be removed

src/RelayClient.ts
```diff
@@ -44,16 +76,206 @@ class RelayClient extends EnvelopingEventEmitter {
...
+      (await this.estimateTokenTransferGas({
+        gasPrice,
+        tokenContract,
+        tokenAmount,
+        callForwarder,
+        isSmartWalletDeploy,
+        preDeploySWAddress,
+      }));
+
+    if (!tokenGas ?? BigNumber.from(tokenGas).isZero()) {
```
franciscotobar, line 201: tokenGas can be zero, the nonZero validation can be removed

src/RelayClient.ts
jurajpiar: ah yea I forgot... tx

src/RelayClient.ts
jurajpiar: okidokie tx

franciscotobar (approved):
LGTM

src/RelayClient.ts
jurajpiar: would that happen prior to calling this method then?

franciscotobar: yes the way that I see it is relayTransaction -> selectNextRelay -> _getEnvelopingRequestDetails -> _prepareRelayHttpRequest.

But Im not pretty sure but when the selectNextRelay is called we will have a server available that give us the feesReceiver, since its the one that will process the transaction, this needs to be different for each server, he question part that I still does not have clear if we should do the _getEnvelopingRequestDetails and _prepareRelayHttpRequest for each request or change just the necessary.

src/RelayClient.ts
```diff
@@ -48,16 +80,188 @@ class RelayClient extends EnvelopingEventEmitter {
...
+    if (!relayServerUrl) {
+      throw new Error(
+        'Check that your configuration contains at least one preferred relay with url.'
+      );
+    }
+    const feesReceiver: PromiseOrValue<string> = (
+      await new HttpClient().getChainInfo(relayServerUrl)
+    ).feesReceiver;
+
+    if (!feesReceiver) {
```
franciscotobar, line 135: Are we going to keep like this or are you going to change it or should I change it once I need it?

jurajpiar: changed now

src/RelayClient.ts
jurajpiar: changed

PR #60, opened 2022-12-12
PP606: migrate RelayClient._estimateTokenTransferGas

AndresQuijano:
## What

- Migrate the public function estimateTokenTransferGas() of the RelayClient to use Ethers.js and Hardhat

## Why

- Project modernization

## Refs

- [PP-606](https://rsklabs.atlassian.net/browse/PP-606)


src/constants/configs.ts
```diff
@@ -1,3 +1,6 @@
 const ROOT = 'EnvelopingConfig';
 
+export const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000';
+export const ESTIMATED_GAS_CORRECTION_FACTOR = '1.0'; // TODO: if needed put a correction factor to mitigate RSK node gas miscalculation if execution includes refunds
+export const INTERNAL_TRANSACTION_ESTIMATE_CORRECTION = '20000'; // When estimating the gas an internal call is going to spend, we need to substract some gas inherent to send the parameters to the blockchain
```
jurajpiar, line 5: 1. This file should probably only contain config-file-related constants, however I'm open to discussion.
3. If the value are numeric let's try to have them numeric, or big number, but not string

src/RelayClient.ts
```diff
@@ -103,6 +111,57 @@ class RelayClient extends EnvelopingEventEmitter {
...
+    internalTransactionEstimateCorrection = INTERNAL_TRANSACTION_ESTIMATE_CORRECTION,
+    estimatedGasCorrectionFactor = ESTIMATED_GAS_CORRECTION_FACTOR
+  ): Promise<BigNumber> {
+    const tokenContract = envelopingRequest.request.tokenContract.toString();
+
+    if (
+      tokenContract === ZERO_ADDRESS ||
+      envelopingRequest.request.tokenAmount === 0
+    ) {
+      BigNumber.from(0);
```
jurajpiar, line 130: fyi `ethers.constants.Zero`

src/constants/configs.ts
```diff
@@ -1,3 +1,6 @@
 const ROOT = 'EnvelopingConfig';
 
+export const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000';
```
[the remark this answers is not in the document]
AndresQuijano: Done.

src/constants/configs.ts
AndresQuijano: I just moved this to utils.ts and changed it to numeric.

src/RelayClient.ts
```diff
@@ -103,6 +111,57 @@ class RelayClient extends EnvelopingEventEmitter {
       metadata,
     };
   };
+
+  // TODO: lift the ts exception once method is used
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  private async _estimateTokenTransferGas(
+    requestConfig: RequestConfig,
```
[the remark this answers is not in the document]
AndresQuijano: As we talked during the daily the types are not the same.

src/RelayClient.ts
AndresQuijano: Thanks. Done.

antomor (changes requested):
@AndresQuijano Thank you for your changes. Apart from the comment I left, I've a couple of doubts:
- why do we call some variable fakes if we use them for real? e.g. `FAKE_INTERNAL_CORRECTION`, `FAKE_CORRECTION_FACTOR`. It doesn't apply to `FAKE_GAS_COST`.
- Why do we need to make a deep copy of the config? do we change it in the method?

src/RelayClient.ts
```diff
@@ -126,15 +141,88 @@ class RelayClient extends EnvelopingEventEmitter {
...
+    const erc20 = IERC20__factory.connect(tokenContract, this._provider);
+    const gasCost = await erc20.estimateGas.transfer(
+      tokenContract,
+      envelopingRequest.request.tokenAmount,
+      { from: tokenOrigin }
+    );
+
+    const internalCallCost = gasCost.gt(internalTransactionEstimateCorrection)
+      ? gasCost.sub(internalTransactionEstimateCorrection)
+      : gasCost;
```
antomor, line 209: If we agree on having a [single function that applies this logic](https://github.com/rsksmart/rif-relay-client/pull/55/files#r1048742211), we could call it here also.

src/RelayClient.ts
```diff
@@ -126,15 +141,88 @@ class RelayClient extends EnvelopingEventEmitter {
...
+        estimatedGasCorrectionFactor
+      );
+      const internalCallCostBN = BignumberJs(internalCallCost.toString());
+
+      return BigNumber.from(
+        internalCallCostBN
+          .multipliedBy(estimatedGasCorrectionFactorBN)
+          .toString()
+      );
+    }
```
antomor, line 222: The same applies here: if we manage to have that logic in a [utility function](https://github.com/rsksmart/rif-relay-client/pull/55/files#r1048744015), we could use it here

antomor (commented):
We could decide to extract those utility functions in this PR or in the other one.

AndresQuijano:
> @AndresQuijano Thank you for your changes. Apart from the comment I left, I've a couple of doubts:
> 
> * why do we call some variable fakes if we use them for real? e.g. `FAKE_INTERNAL_CORRECTION`, `FAKE_CORRECTION_FACTOR`. It doesn't apply to `FAKE_GAS_COST`.
Well, I tend to use the fake word before everything that I use in testing. It's just a naming thing but we can discuss about it.

> * Why do we need to make a deep copy of the config? do we change it in the method?
Yes, I was changing the object and it was making fail other tests. The other option is to keep the original value and put it back after the test but I prefer this one.

test/RelayClient.test.ts
```diff
@@ -361,14 +373,169 @@ describe('RelayClient', function () {
...
+        envelopingRequest: EnvelopingRequest,
+        internalTransactionEstimateCorrection?: number,
+        estimatedGasCorrectionFactor?: number
+      ) => Promise<BigNumber>;
+    };
+
+    let ierc20TransferStub: SinonStub;
+
+    beforeEach(function () {
+      relayClient = new RelayClient() as unknown as RelayClientExposed;
```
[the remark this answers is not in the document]
AndresQuijano: You're right. I did it this way because initially the method was accorded to be private. I just changed it.

test/RelayClient.test.ts
```diff
@@ -361,14 +373,169 @@ describe('RelayClient', function () {
...
+      relayRequest.request.tokenAmount = BigNumber.from(0);
+
+      expect(
+        (
+          await relayClient.estimateTokenTransferGas({}, relayRequest)
+        ).toString()
+      ).to.be.equal(BigNumber.from(0).toString());
+    });
+
+    it('Shold fail if it is a deploy and the smartWallet is missing', async function () {
```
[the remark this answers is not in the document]
AndresQuijano: Thanks.

test/RelayClient.test.ts
```diff
@@ -361,14 +373,169 @@ describe('RelayClient', function () {
...
+        FAKE_REQUEST_CONFIG
+      ) as RequestConfig;
+      requestConfig.smartWalletAddress = constants.AddressZero;
+
+      await expect(
+        relayClient.estimateTokenTransferGas(requestConfig, FAKE_DEPLOY_REQUEST)
+      ).to.be.rejectedWith(MISSING_SMART_WALLET_ADDRESS);
+    });
+
+    it("Shold fail if it's a relay transaction and the callForwarder is missing", async function () {
```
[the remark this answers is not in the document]
AndresQuijano: Thanks.

src/RelayClient.ts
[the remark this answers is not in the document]
jurajpiar: yes, my bad

src/RelayClient.ts
```diff
@@ -126,15 +141,88 @@ class RelayClient extends EnvelopingEventEmitter {
...
       return false;
     }
 
     return true;
   }
-  
+
+  // TODO: lift the ts exception once method is used
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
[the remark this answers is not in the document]
AndresQuijano: Done

test/RelayClient.test.ts
```diff
@@ -361,14 +372,160 @@ describe('RelayClient', function () {
...
+          await relayClient.estimateTokenTransferGas({}, relayRequest)
+        ).toString()
+      ).to.be.equal(BigNumber.from(0).toString());
+    });
+
+    it('Should fail if it is a deploy and the smartWallet is missing', async function () {
+      const requestConfig = config.util.toObject(
+        FAKE_REQUEST_CONFIG
+      ) as RequestConfig;
+      requestConfig.smartWalletAddress = constants.AddressZero;
```
franciscotobar, line 443: Im not see where are we validating or testing if the smartWalletAddress is missing, since we are assigning a AddressZero, I remove this and the test fails, meaning that the smartWalletAddress can be empty and does not work as expected

src/RelayClient.ts
```diff
@@ -126,15 +141,85 @@ class RelayClient extends EnvelopingEventEmitter {
...
+      tokenContract === constants.AddressZero ||
+      envelopingRequest.request.tokenAmount.toString() === '0'
+    ) {
+      return ethers.constants.Zero;
+    }
+
+    let tokenOrigin: string | undefined;
+
+    if (requestConfig.isSmartWalletDeploy) {
+      tokenOrigin = requestConfig?.smartWalletAddress;
```
franciscotobar, line 181:
```suggestion
      tokenOrigin = requestConfig.smartWalletAddress;
```

src/RelayClient.ts
```diff
@@ -126,15 +141,85 @@ class RelayClient extends EnvelopingEventEmitter {
...
+      tokenOrigin = envelopingRequest.relayData.callForwarder.toString();
+
+      if (tokenOrigin === constants.AddressZero) {
+        throw Error(MISSING_CALL_FORWARDER);
+      }
+    }
+
+    const erc20 = IERC20__factory.connect(tokenContract, this._provider);
+    const gasCost = await erc20.estimateGas.transfer(
+      tokenContract,
```
franciscotobar, line 199: This value is not the token contract, is the feesReceiver address.

src/RelayClient.ts
AndresQuijano: Done.

test/RelayClient.test.ts
AndresQuijano: Done.

src/RelayClient.ts
AndresQuijano: Done.

jurajpiar:
> why do we call some variable fakes if we use them for real? e.g. FAKE_INTERNAL_CORRECTION, FAKE_CORRECTION_FACTOR. It doesn't apply to FAKE_GAS_COST.

@AndresQuijano I think such generic approach may be damaging in long term, as makes things less obvious. It would be better if we use:
- `fake` for faking values
- `stub` for stubbing functionality
- `mock` for mocking the object under test
- `expected` for values that the test expects
- `actual` for collected values from the function under test
- and no prefix/suffix at all for everything else.

AndresQuijano:
> > > why do we call some variable fakes if we use them for real? e.g. FAKE_INTERNAL_CORRECTION, FAKE_CORRECTION_FACTOR. It doesn't apply to FAKE_GAS_COST.
> 
> > Well, I tend to use the fake word before everything that I use in testing. It's just a naming thing but we can discuss about it.
> 
> @AndresQuijano I think such generic approach may be damaging in long term, as makes things less obvious. It would be better if we use:
> 
> * `fake` for faking values
> * `stub` for stubbing functionality
> * `mock` for mocking the object under test
> * `expected` for values that the test expects
> * `actual` for collected values from the function under test
> * and no prefix/suffix at all for everything else.

Thanks. I just changed the naming style.

antomor (approved):

PR #62, opened 2022-12-15
PP601: Port the GasEstimator from the server to the client

AndresQuijano:
## What

- Port the GasEstimator from rif-relay-server to rif-relay-client

## Why

- Project refactoring

## Refs

- [PP-601](https://rsklabs.atlassian.net/browse/PP-601)

### DIFF
--- a/src/RelayClient.ts
+++ b/src/RelayClient.ts
@@ -56,6 +56,7 @@ type RequestConfig = {
   estimatedGasCorrectionFactor?: BigNumberish;
 };
 
+//FIXME name standardization
 type TokenGasEstimationParams = Pick<
   EnvelopingRequest['request'],
   'tokenAmount' | 'tokenContract'
@@ -72,6 +73,7 @@ type TokenGasEstimationParams = Pick<
     | 'estimatedGasCorrectionFactor'
   >;
 
+//FIXME name standardization
 type EstimateInternalGasParams = Pick<
   RelayRequestBody,
   'data' | 'to' | 'from'
@@ -348,11 +350,11 @@ class RelayClient extends EnvelopingEventEmitter {
     );
   }
 
-  public estimateInternalCallGas = async ({
+  public async estimateInternalCallGas({
     internalEstimationCorrection,
     estimatedGasCorrectionFactor,
     ...estimateGasParams
-  }: EstimateInternalGasParams): Promise<BigNumber> => {
+  }: EstimateInternalGasParams): Promise<BigNumber> {
     let estimation: BigNumber = await this._provider.estimateGas(
       estimateGasParams
     );
@@ -363,7 +365,7 @@ class RelayClient extends EnvelopingEventEmitter {
     );
 
     return applyGasCorrectionFactor(estimation, estimatedGasCorrectionFactor);
-  };
+  }
 
   public async estimateTokenTransferGas({
     internalEstimationCorrection,
--- a/src/gasEstimator/gasEstimator.ts
+++ b/src/gasEstimator/gasEstimator.ts
@@ -1,0 +1,50 @@
+import type { BigNumber } from 'ethers';
+import { isDeployRequest } from '../common/relayRequest.utils';
+import type { EnvelopingTxRequest } from '../common/relayTransaction.types';
+import RelayClient from '../RelayClient';
+import {
+  standardMaxPossibleGasEstimation,
+  linearFitMaxPossibleGasEstimation,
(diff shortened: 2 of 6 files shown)


franciscotobar:
Need to merge https://github.com/rsksmart/rif-relay-client/pull/60 and rebase.

src/gasEstimator/gasEstimator.ts
```diff
@@ -0,0 +1,50 @@
...
+): Promise<BigNumber> => {
+  const {
+    relayRequest,
+    metadata: { signature },
+  } = request;
+
+  const { tokenContract, tokenAmount } = relayRequest.request;
+  const { feesReceiver, callForwarder, gasPrice } = relayRequest.relayData;
+
+  const relayClient = new RelayClient();
```
franciscotobar, line 22: By doing this here I think that we are creating a dependency on the relayClient and the constructor change, we will need to include those changes here or even pass the client as parameter.

src/gasEstimator/utils.ts
```diff
@@ -0,0 +1,122 @@
...
+  if (isDeployRequest(envelopingRequest)) {
+    throw Error('LinearFit estimation not implemented for deployments');
+  }
+
+  const {
+    request: { data, from, to },
+    relayData: { gasPrice },
+  } = envelopingRequest;
+
+  const relayClient = new RelayClient();
```
franciscotobar, line 64: By doing this here I think that we are creating a dependency on the relayClient and the constructor change, we will need to include those changes here or even pass the client as parameter.

src/gasEstimator/gasEstimator.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const isSmartWalletDeploy = isDeployRequest(relayRequest);
+
+  //FIXME validate how to generate the smart wallet address
+  const preDeploySWAddress = isSmartWalletDeploy ? undefined : undefined;
```
franciscotobar, line 27: To generate the SW Address I could generate my own function or use https://github.com/rsksmart/rif-relay-client/blob/PP601_gasEstimator/src/discovery/utils.ts#L147, but this adds more complexity for a small task.

src/RelayClient.ts
```diff
@@ -53,19 +56,33 @@ type RequestConfig = {
   estimatedGasCorrectionFactor?: BigNumberish;
 };
 
+//FIXME name standardization
 type TokenGasEstimationParams = Pick<
```
antomor, line 60: IMHO having the types defined in this way makes the code very difficult to read.
No need to change it now, but we may probably discuss it with the rest of the team

src/RelayClient.ts
```diff
@@ -53,19 +56,33 @@ type RequestConfig = {
...
+  Pick<
+    RequestConfig,
+    | 'isSmartWalletDeploy'
+    | 'preDeploySWAddress'
+    | 'internalEstimationCorrection'
+    | 'estimatedGasCorrectionFactor'
+  >;
+
+//FIXME name standardization
 type EstimateInternalGasParams = Pick<
```
antomor, line 77: IMHO having the types defined in this way makes the code very difficult to read.
No need to change it now, but we may probably discuss it with the rest of the team

src/gasEstimator/gasEstimator.ts
antomor: Does this mean we're going to have a circular dependency?

src/gasEstimator/gasEstimator.ts
antomor: totally agree, we may consider this improvement in future.

src/gasEstimator/utils.ts
antomor: does this mean we're going to introduce a circular dependency?

antomor (approved):
@franciscotobar thank you for the changes, I've just left doubts, no changes strictly required.

src/gasEstimator/gasEstimator.ts
franciscotobar: not circular but we are coupling this method with the relayClient.

src/gasEstimator/utils.ts
franciscotobar: not circular but we are coupling this method with the relayClient.

antomor: I think we can simplify the interaction later on. I'd leave it as it is for now

PR #63, opened 2022-12-20
PP-498 relay provider

ironFe93:
## What

Change RelayProvider.ts to ethers version

## Why

Part of a general upgrade

## Refs
 https://rsklabs.atlassian.net/jira/software/projects/PP/boards/84/backlog?selectedIssue=PP-498

### DIFF
--- a/src/RelayProvider.ts
+++ b/src/RelayProvider.ts
@@ -1,14 +1,15 @@
-import { constants, Transaction } from 'ethers';
+import type { Transaction } from 'ethers';
 import type { Networkish } from "@ethersproject/networks";
 import { JsonRpcProvider, TransactionReceipt } from "@ethersproject/providers";
 import type { ConnectionInfo } from "@ethersproject/web";
 import { getEnvelopingConfig } from './utils';
-import { hexValue } from "@ethersproject/bytes";
 import RelayClient, { RequestConfig } from './RelayClient';
 import log from 'loglevel';
 import { RelayHub__factory } from '@rsksmart/rif-relay-contracts';
 import AccountManager from './AccountManager';
 import type { UserDefinedEnvelopingRequest } from './common/relayRequest.types';
+import { useEnveloping } from './utils';
+import BigNumber from 'bignumber.js';
 
 export interface RelayingResult {
     validUntilTime?: string;
@@ -18,11 +19,8 @@ export interface RelayingResult {
 
 export default class RelayProvider extends JsonRpcProvider {
 
-    //TODO: remove & when relayTransaction PR is merged and rebased
-    private readonly relayClient!: RelayClient & { relayTransaction: (
-        envelopingRequest: UserDefinedEnvelopingRequest,
-        requestConfig: RequestConfig) => Promise<Transaction> };
-    
+    private readonly relayClient!: RelayClient;
+
     private readonly jsonRpcProvider: JsonRpcProvider;
 
     constructor(
@@ -32,28 +30,7 @@ export default class RelayProvider extends JsonRpcProvider {
         jsonRpcProvider = new JsonRpcProvider(url, network)){
         super(url, network);
         this.jsonRpcProvider = jsonRpcProvider;
-        //@ts-ignore
         this.relayClient = relayClient
-    }
-
-    //I am aware this sucks
-    private _useEnveloping(method: string, params: Array<Record<string, unknown>>): boolean {
-        if(method === 'eth_accounts'){
-            return true;
-        }
-
-        const [a] = params;
-
-        if(!a){
-            return false;
-        }
-
-        const { envelopingTx, requestConfig } = a; 
-        if(envelopingTx && requestConfig && (requestConfig as RequestConfig).useEnveloping){
-            return true;
-        }
-
-        return false;
     }
 
     private _getRelayStatus(respResult: TransactionReceipt): {
@@ -76,11 +53,7 @@ export default class RelayProvider extends JsonRpcProvider {
 
         if (recipientRejectedEvents) {
             const { reason } = recipientRejectedEvents.args;
-
-            const revertReason = (reason as string) ?? 'Unknown';
-            log.info(
-                `Recipient rejected on-chain: ${revertReason}`
-            );
+            const revertReason = (reason as string);
 
             return {
                 relayRevertedOnRecipient: true,
@@ -98,24 +71,14 @@ export default class RelayProvider extends JsonRpcProvider {
             };
         }
 
-        // Check if it wasn't a deploy call which does not emit TransactionRelayed events but Deployed events
-        const deployedEvent = parsedLogs.find(log => log.name == 'Deployed');
-
-        if (deployedEvent) {
-            return {
-                relayRevertedOnRecipient: false,
-                transactionRelayed: true
-            };
-        }
-
         log.info(
-            'Neither TransactionRelayed, Deployed, nor TransactionRelayedButRevertedByRecipient events found. This might be a non-enveloping transaction '
+            'Neither TransactionRelayed, nor TransactionRelayedButRevertedByRecipient events found. This might be a non-enveloping transaction.'
         );
 
         return {
             relayRevertedOnRecipient: false,
             transactionRelayed: false,
-            reason: 'Neither TransactionRelayed, Deployed, nor TransactionRelayedButRevertedByRecipient events found. This might be a non-enveloping transaction'
+            reason: 'Neither TransactionRelayed, Deployed, nor TransactionRelayedButRevertedByRecipient events found. This might be a non-enveloping transaction.'
         };
     }
 
@@ -137,7 +100,7 @@ export default class RelayProvider extends JsonRpcProvider {
 
             let receipt;
             if(!requestConfig.ignoreTransactionReceipt){
-                receipt = await this.getTransactionReceipt(hash);
+                receipt = await this.waitForTransaction(hash, 1);
             }
 
             return {
@@ -165,31 +128,20 @@ export default class RelayProvider extends JsonRpcProvider {
         let onlyPreferredRelaysValue = requestConfig.onlyPreferredRelays;
         let gasToSend = requestConfig.forceGasLimit;
 
-        if (
-            callForwarderValue === undefined ||
-            callForwarderValue === null ||
-            callForwarderValue === constants.AddressZero
-        ) {
+        if (!Number(callForwarderValue)) {
             callForwarderValue = config.forwarderAddress;
         }
 
-        if (
-            relayHubValue === undefined ||
-            relayHubValue === null ||
-            relayHubValue === constants.AddressZero
-        ) {
+        if (!Number(relayHubValue)) {
             relayHubValue = config.relayHubAddress;
         }
 
-        if (
-            onlyPreferredRelaysValue === undefined ||
-            onlyPreferredRelaysValue === null
-        ) {
+        if (!onlyPreferredRelaysValue) {
             onlyPreferredRelaysValue = config.onlyPreferredRelays;
         }
 
-        if (gasToSend !== undefined && gasToSend !== null) {
-            gasToSend = hexValue(BigInt(gasToSend));
+        if (gasToSend) {
+            gasToSend = BigNumber(gasToSend).toString();
         }
 
         /**
@@ -202,7 +154,7 @@ export default class RelayProvider extends JsonRpcProvider {
          * value that comes from the original provider
          */
 
-        const fullEnvelopingRequest: UserDefinedEnvelopingRequest = {
+        const userDefinedEnvelopingReq: UserDefinedEnvelopingRequest = {
             relayData: {
                 ...envelopingRequest.relayData,
                 callForwarder: callForwarderValue,
@@ -219,7 +171,7 @@ export default class RelayProvider extends JsonRpcProvider {
             onlyPreferredRelays: onlyPreferredRelaysValue
         }
 
-        const relayingResult = await this.executeRelayTransaction(fullEnvelopingRequest, fullRequestConfig);
+        const relayingResult = await this.executeRelayTransaction(userDefinedEnvelopingReq, fullRequestConfig);
 
         if(relayingResult.receipt){
             const { relayRevertedOnRecipient, reason } = this._getRelayStatus(relayingResult.receipt);
@@ -227,7 +179,7 @@ export default class RelayProvider extends JsonRpcProvider {
 
             if (relayRevertedOnRecipient) {
                 throw new Error(
-                        `Transaction Relayed but reverted on recipient - TxHash: ${transactionHash} , Reason: ${reason ?? 'unknown'}`
+                        `Transaction Relayed but reverted on recipient - TxHash: ${transactionHash} , Reason: ${reason as string}`
                 );
             }
         }
@@ -252,7 +204,7 @@ export default class RelayProvider extends JsonRpcProvider {
     }
 
     override send(method: string, params: Array<Record<string, unknown>>): Promise<unknown> {
-        if (this._useEnveloping(method, params)) {
+        if (useEnveloping(method, params)) {
             const { requestConfig, envelopingTx } = params[0] as { requestConfig: RequestConfig, envelopingTx: UserDefinedEnvelopingRequest }
             const { request } = envelopingTx;
             if (method === 'eth_sendTransaction') {
--- a/src/RelayProvider.ts
+++ b/src/RelayProvider.ts
@@ -1,0 +1,3 @@
+export default class RelayProvider{
+    
+}
--- a/src/RelayProvider.ts
+++ b/src/RelayProvider.ts
@@ -189,7 +189,7 @@ export default class RelayProvider extends JsonRpcProvider {
         }
 
         if (gasToSend !== undefined && gasToSend !== null) {
-            gasToSend = hexValue(gasToSend);
+            gasToSend = hexValue(BigInt(gasToSend));
         }
 
         /**
@@ -236,19 +236,19 @@ export default class RelayProvider extends JsonRpcProvider {
     }
 
     override async listAccounts() {
-        let response = await this.jsonRpcProvider.listAccounts();
+        let accountsList = await this.jsonRpcProvider.listAccounts();
 
-        if (response != null && Array.isArray(response)) {
+        if (accountsList && Array.isArray(accountsList)) {
             const { chainId } = await this.getNetwork();
             const accountManager = new AccountManager(this, chainId);
 
             const ephemeralAccounts =
                 accountManager.getAccounts();
 
-            response = response.concat(ephemeralAccounts);
+            accountsList = accountsList.concat(ephemeralAccounts);
         }
 
-        return response;
+        return accountsList;
     }
 
     override send(method: string, params: Array<Record<string, unknown>>): Promise<unknown> {
--- a/src/RelayProvider.ts
+++ b/src/RelayProvider.ts
@@ -1,4 +1,4 @@
-import type { Transaction } from 'ethers';
+import { Transaction, BigNumber } from 'ethers';
 import type { Networkish } from "@ethersproject/networks";
 import { JsonRpcProvider, TransactionReceipt } from "@ethersproject/providers";
 import type { ConnectionInfo } from "@ethersproject/web";
@@ -9,7 +9,6 @@ import { RelayHub__factory } from '@rsksmart/rif-relay-contracts';
 import AccountManager from './AccountManager';
 import type { UserDefinedEnvelopingRequest } from './common/relayRequest.types';
 import { useEnveloping } from './utils';
-import BigNumber from 'bignumber.js';
 import type { RelayHubInterface } from '@rsksmart/rif-relay-contracts/dist/typechain-types/contracts/RelayHub';
 import type { LogDescription } from 'ethers/lib/utils';
 import type { Either } from './common/utility.types';
@@ -144,7 +143,7 @@ export default class RelayProvider extends JsonRpcProvider {
         }
 
         if (gasToSend) {
-            gasToSend = BigNumber(gasToSend).toString();
+            gasToSend = BigNumber.from(gasToSend).toString();
         }
 
         /**
--- a/src/RelayProvider.ts
+++ b/src/RelayProvider.ts
@@ -11,6 +11,7 @@ import type { UserDefinedEnvelopingRequest } from './common/relayRequest.types';
 import { useEnveloping } from './utils';
 import BigNumber from 'bignumber.js';
 import type { RelayHubInterface } from '@rsksmart/rif-relay-contracts/dist/typechain-types/contracts/RelayHub';
+import type { LogDescription } from 'ethers/lib/utils';
 
 export interface RelayingResult {
     validUntilTime?: string;
@@ -43,8 +44,8 @@ export default class RelayProvider extends JsonRpcProvider {
                 const { url, network } = providerParams;
 
                 super(url, network);
-                new JsonRpcProvider(url, network);
-
+                this.jsonRpcProvider = new JsonRpcProvider(url, network);
+                
             }
             this.relayClient = relayClient;
     }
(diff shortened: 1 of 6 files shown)


src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+        super(url, network);
+        this.jsonRpcProvider = new JsonRpcProvider(url, network);
+        //TODO: remove this when relayTransaction PR is merged and rebased
+        //@ts-ignore
+        this.relayClient =
+            relayClient ??
+            new RelayClient();
+    }
+
+    private _useEnveloping(method: string, requestConfig: RequestConfig): boolean {
```
jurajpiar, line 38: can be refactored out as a utility fn

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+    override send(method: string, params: Array<{ requestConfig: RequestConfig, envelopingTx: UserDefinedEnvelopingRequest }>): Promise<unknown> {
+
+        const [ envelopingData ] = params;
+        if(!envelopingData) {
+            throw new Error('No data to send!');
+        }
+
+        const { requestConfig, envelopingTx: { request } } = envelopingData;
+
+        if (this._useEnveloping(method, requestConfig)) {
```
jurajpiar, line 245: considering that this method is only called here and performs the same check as on line 261, it seems superfluous. Probably could simplify this.

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+export default class RelayProvider extends JsonRpcProvider {
+
+    //TODO: remove & when relayTransaction PR is merged and rebased
+    private readonly relayClient!: RelayClient & { relayTransaction: (
+        envelopingRequest: UserDefinedEnvelopingRequest,
+        requestConfig: RequestConfig) => Promise<Transaction> };
+    
+    private readonly jsonRpcProvider: JsonRpcProvider;
+
+    constructor(url?: ConnectionInfo | string, network?: Networkish, relayClient?: RelayClient){
```
antomor, line 28: I'd allow the user to specify either the JsonRpcProvider or the parameters required to build it.

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    override send(method: string, params: Array<{ requestConfig: RequestConfig, envelopingTx: UserDefinedEnvelopingRequest }>): Promise<unknown> {
+
+        const [ envelopingData ] = params;
+        if(!envelopingData) {
+            throw new Error('No data to send!');
+        }
+
+        const { requestConfig, envelopingTx: { request } } = envelopingData;
```
antomor, line 243: I'm not sure that this is going to work for transactions that are not intended to be relayed. If we try to extract properties from undefined properties, it raises an error, and I believe that neither `envelopingData` or `envelopingTx` is going to be defined in that case.

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+        requestConfig: RequestConfig
+    ): Promise<RelayingResult> {
+        const config = getEnvelopingConfig();
+
+        const { relayData, request } = envelopingRequest;
+
+        let callForwarderValue = relayData.callForwarder;
+        let relayHubValue = request.relayHub;
+        let onlyPreferredRelaysValue = requestConfig.onlyPreferredRelays;
+        let gasToSend = requestConfig.forceGasPrice;
```
antomor, line 160: I think we should use `forceGasLimit` instead

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+            const transaction = await this.relayClient.relayTransaction(
+                envelopingRequest, requestConfig
+            );
+
+            const { hash } = transaction;
+
+            if(!hash){
+                throw new Error('Transaction has no hash!');
+            }
+            const receipt = await this.getTransactionReceipt(hash);
```
antomor, line 127: Does the JsonRpcProvider always wait for the transactionReceipt to be ready? If not, I don't think we should do the same here

antomor: We could eventually add an optional parameter to let the user decide what to do.

antomor (changes requested):
I've some doubts about the JsonRpcProvider behaviour in ethers.js, so I'd try to reproduce the exact same behaviour if possible, otherwise it's difficult to use it transparently

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+            return false;
+        }
+
+        return requestConfig.useEnveloping ?? false;
+    }
+
+    private _getRelayStatus(respResult: TransactionReceipt): {
+        relayRevertedOnRecipient: boolean;
+        transactionRelayed: boolean;
+        reason?: string;
```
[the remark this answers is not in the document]
ironFe93: Used on line 119

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+            };
+        }
+
+        const relayHubInterface = RelayHub__factory.createInterface();
+        const parsedLogs = respResult.logs.map(log => relayHubInterface.parseLog(log));
+
+        const recipientRejectedEvents = parsedLogs.find(log => log.name == 'TransactionRelayedButRevertedByRecipient');
+
+        if (recipientRejectedEvents) {
+            const { reason } = recipientRejectedEvents.args;
```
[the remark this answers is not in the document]
ironFe93: uhh I don't see the shadowed constant, could you be more specific?

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+
+    private async _ethSendTransaction(
+        envelopingRequest: UserDefinedEnvelopingRequest,
+        requestConfig: RequestConfig
+    ): Promise<RelayingResult> {
+        const config = getEnvelopingConfig();
+
+        const { relayData, request } = envelopingRequest;
+
+        let callForwarderValue = relayData.callForwarder;
```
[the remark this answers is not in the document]
ironFe93: true, thanks!

src/RelayProvider.ts
ironFe93: got it

src/RelayProvider.ts
ironFe93: got it

src/RelayProvider.ts
ironFe93: Actually, the provider used to have this functionality, I just missed it so I will be adding it back.

src/RelayProvider.ts
ironFe93: I have changed the function signature to be more flexible

src/RelayProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+        return accountsList;
+    }
+
+    override send(method: string, params: Array<Record<string, unknown>>): Promise<unknown> {
+        if (this._useEnveloping(method, params)) {
+            const { requestConfig, envelopingTx } = params[0] as { requestConfig: RequestConfig, envelopingTx: UserDefinedEnvelopingRequest }
+            const { request } = envelopingTx;
+            if (method === 'eth_sendTransaction') {
+
+                if (!request.to) {
```
antomor, line 260: Are we sure `to` can be found in the request object? According to the [previous code](https://github.com/rsksmart/rif-relay-client/blob/de33ebb63c57d6b7a94071baf480fbbc958aa790/src/RelayProvider.ts#L124), `to` was retrieved from `params[0]` directly and not from an inner property.

src/RelayProvider.ts
antomor: We reverted this to wait for transaction request only if `ignoreTransactionReceipt` is false

src/RelayProvider.ts
antomor: I meant we should either accept `url` and `network` or a JsonRpcProvider.
If JsonRpcProvider is received, we could extract `url` from `provider.connection.url` and `network` from `provider.network`

src/RelayProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+
+            const { hash } = transaction;
+
+            if(!hash){
+                throw new Error('Transaction has no hash!');
+            }
+
+            let receipt;
+            if(!requestConfig.ignoreTransactionReceipt){
+                receipt = await this.getTransactionReceipt(hash);
```
franciscotobar, line 140: I think that ethers has a special method, to wait for the transaction. 
https://docs.ethers.org/v5/api/providers/provider/#Provider-waitForTransaction

src/RelayProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+
+        if (
+            onlyPreferredRelaysValue === undefined ||
+            onlyPreferredRelaysValue === null
+        ) {
+            onlyPreferredRelaysValue = config.onlyPreferredRelays;
+        }
+
+        if (gasToSend !== undefined && gasToSend !== null) {
+            gasToSend = hexValue(BigInt(gasToSend));
```
franciscotobar, line 192: I think that we should use a BigNumber here

src/RelayProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+        let relayHubValue = await request.relayHub;
+        let onlyPreferredRelaysValue = requestConfig.onlyPreferredRelays;
+        let gasToSend = requestConfig.forceGasLimit;
+
+        if (
+            callForwarderValue === undefined ||
+            callForwarderValue === null ||
+            callForwarderValue === constants.AddressZero
+        ) {
+            callForwarderValue = config.forwarderAddress;
```
franciscotobar, line 173: The only reason that I see for having a `callForwarder` in the config, would be for deployments but we already have the `smartWalletFactoryAddress`, unless that Im missing something.

src/RelayProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    override send(method: string, params: Array<Record<string, unknown>>): Promise<unknown> {
+        if (this._useEnveloping(method, params)) {
+            const { requestConfig, envelopingTx } = params[0] as { requestConfig: RequestConfig, envelopingTx: UserDefinedEnvelopingRequest }
+            const { request } = envelopingTx;
+            if (method === 'eth_sendTransaction') {
+
+                if (!request.to) {
+                    throw new Error(
+                        'Relay Provider cannot relay contract deployment transactions. Add {from: accountWithRBTC, useEnveloping: false}.'
```
franciscotobar, line 262: Im not sure about this, on smart wallet deployments, the to can be undefined.

src/RelayProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+                }
+
+                return this._ethSendTransaction(envelopingTx, requestConfig);
+            }
+
+            if (method === 'eth_accounts') {
+                return this.listAccounts();
+            }
+
+            throw new Error(`Rif Relay unsupported method: ${ method }`);
```
franciscotobar, line 273: By having this throw Error, are we not disabling methods than can executed without the need of enveloping or any change of behavior?

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+
+        const relayHubInterface = RelayHub__factory.createInterface();
+        const parsedLogs = respResult.logs.map(log => relayHubInterface.parseLog(log));
+
+        const recipientRejectedEvents = parsedLogs.find(log => log.name == 'TransactionRelayedButRevertedByRecipient');
+
+        if (recipientRejectedEvents) {
+            const { reason } = recipientRejectedEvents.args;
+
+            const revertReason = (reason as string) ?? 'Unknown';
```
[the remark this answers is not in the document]
ironFe93: Leaving it as is instead of specifying 'Unknown'

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+        if (!respResult.logs || respResult.logs.length === 0) {
+            return {
+                relayRevertedOnRecipient: false,
+                transactionRelayed: false,
+                reason: 'Tx logs not found'
+            };
+        }
+
+        const relayHubInterface = RelayHub__factory.createInterface();
+        const parsedLogs = respResult.logs.map(log => relayHubInterface.parseLog(log));
```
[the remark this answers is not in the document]
ironFe93: This type is not exported on the contracts repo, I would have to import it by going all the way into the dist folder and I don't think this is desirable.

test/RelayProvider.test.ts
```diff
@@ -0,0 +1,73 @@
+import { JsonRpcProvider } from "@ethersproject/providers";
+import { expect, use } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import config from 'config';
+import Sinon from 'sinon';
+import sinonChai from 'sinon-chai';
+import RelayProvider from '../src/RelayProvider';
+import RelayClient from 'src/RelayClient';
```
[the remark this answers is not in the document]
ironFe93: fixed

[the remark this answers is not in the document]
ironFe93: fixed

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+
+        if (transactionRelayed) {
+            return {
+                relayRevertedOnRecipient: false,
+                transactionRelayed: true
+            };
+        }
+
+        // Check if it wasn't a deploy call which does not emit TransactionRelayed events but Deployed events
+        const deployedEvent = parsedLogs.find(log => log.name == 'Deployed');
```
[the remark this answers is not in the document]
ironFe93: This is true, why did this part even exist I have no idea. Removing it.

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+            if(!hash){
+                throw new Error('Transaction has no hash!');
+            }
+            const receipt = await this.getTransactionReceipt(hash);
+
+            const { relayRevertedOnRecipient, reason } = this._getRelayStatus(receipt);
+
+            if (relayRevertedOnRecipient) {
+                throw new Error(
+                        `Transaction Relayed but reverted on recipient - TxHash: ${hash} , Reason: ${reason ?? 'unknown'}`
```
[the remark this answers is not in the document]
ironFe93: replacing with ${reason as string} because the linter does not let me fill a template with possible undefined values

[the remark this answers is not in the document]
ironFe93: true, removed the log in getRelayStatus

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+
+        let callForwarderValue = relayData.callForwarder;
+        let relayHubValue = request.relayHub;
+        let onlyPreferredRelaysValue = requestConfig.onlyPreferredRelays;
+        let gasToSend = requestConfig.forceGasPrice;
+
+        if (
+            callForwarderValue === undefined ||
+            callForwarderValue === null ||
+            callForwarderValue === constants.AddressZero
```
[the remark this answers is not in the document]
ironFe93: yes thank you

src/RelayProvider.ts
ironFe93: true, done

src/RelayProvider.ts
ironFe93: very nice find.

src/RelayProvider.ts
ironFe93: At this point I am unsure about this one. We can discuss further with @antomor

src/RelayProvider.ts
ironFe93: you would have to set the useEnveloping flag to true in the request config to get to this error. To not use the custom functionality, one should set useEnveloping to false. At least this is what I understand as the intended use case.

src/RelayProvider.ts
ironFe93: that is how it used to be, yes. Now however, relay transactions need two parameters, the data itself (in the form of UserDefinedEnvelopingRequest) and the requestConfig object. This means that whoever is implementing the provider to carry out relay transactions must send an array of 
`{ requestConfig: RequestConfig, envelopingTx: UserDefinedEnvelopingRequest }` 
and set the useEnveloping flag in requestConfig to true.

src/RelayProvider.ts
ironFe93: From what I understood, the send method of the relayProvider should not be used to deploy smart wallets.

src/RelayProvider.ts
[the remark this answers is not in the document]
jurajpiar: It is desirable to define the type, otherwise the variable can represent anything. The having to go to dist should have been mitigated, but I seem to remember that it didn't work or something, so it's just something we'll need to resolve sooner or later, def not a reason to not use the generated types from typechain.

src/RelayProvider.ts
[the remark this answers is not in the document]
jurajpiar: aaah haha sorry my bad that was a return type I was looking at. oops

src/RelayProvider.ts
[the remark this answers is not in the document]
jurajpiar: not sure why I didn't see it. my bad

src/RelayProvider.ts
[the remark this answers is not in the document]
franciscotobar: I dont understand, which one is the one that does not have the typing? We have an index in the contracts where we can get or export any desired type from typechain.

src/RelayProvider.ts
franciscotobar: did you push the changes?

ironFe93: I did now

src/RelayProvider.ts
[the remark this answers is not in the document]
ironFe93: `import type { RelayHubInterface } from '@rsksmart/rif-relay-contracts/dist/typechain-types/contracts/RelayHub';
`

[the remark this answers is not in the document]
franciscotobar: I can add it in the rif-relay-contracts. 

```
export {
  EnvelopingTypes,
  StakeUnlockedEvent,
  IForwarder as IForwarderTypes,
  RelayHubInterface
} from './typechain-types/contracts/RelayHub';

File: src/RelayProvider.ts (hunk 3 above)
Comment (ironFe93) [reply src/RelayProvider.ts]: I have just pushed a change for the constructor, let me know if it is what you meant

File: src/RelayProvider.ts [hunk 22]
```diff
@@ -0,0 +1,270 @@
...
+
+        log.info(
+            'Neither TransactionRelayed, Deployed, nor TransactionRelayedButRevertedByRecipient events found. This might be a non-enveloping transaction '
+        );
+
+        return {
+            relayRevertedOnRecipient: false,
+            transactionRelayed: false,
+            reason: 'Neither TransactionRelayed, Deployed, nor TransactionRelayedButRevertedByRecipient events found. This might be a non-enveloping transaction'
+        };
```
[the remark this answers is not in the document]
jurajpiar:
```typescript
const handlers = {
  Deployed: () => ({
                relayRevertedOnRecipient: false,
                transactionRelayed: true
   }),
  TransactionRelayed: () => {}
}

const eventRunner = (eventName: EnvelopingEvents, args) => {

  return handlers[eventName](args);
}
```

src/RelayProvider.ts
antomor: Ok, although I'm not sure this is compatible with a normal request, I got it.

src/RelayProvider.ts
antomor: Yes, indeed. See the [previous code](https://github.com/rsksmart/rif-relay-client/blob/main/src/RelayProvider.ts#L125).

src/RelayProvider.ts
antomor: @ironFe93 I think that by throwing an error we're assuming that the inner provider isn't able to relay the transaction. Although it makes sense, I'd leave the inner provider to raise the exception instead of raising it directly.

src/RelayProvider.ts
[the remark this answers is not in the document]
antomor: @jurajpiar I like it! 👆🏽

src/RelayProvider.ts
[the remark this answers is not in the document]
antomor: I know it's matter of opinion, but I'd prefer to have a managed string `unknown` instead of `undefined`

src/RelayProvider.ts
antomor: Actually I was thinking about something like:
```
constructor(
        relayClient = new RelayClient(),
        providerOrProviderParams: Either<JsonRpcProvider,ProviderParams>){
            let url:  ConnectionInfo | string;
            let network: Networkish;

            if (providerOrProviderParams.connection) {
                url = providerOrProviderParams.connection.url;
                network = providerOrProviderParams._network;
            } else {
                url = providerOrProviderParams.url;
                network = providerOrProviderParams.network
            }
            super(url, network);
            this.relayClient = relayClient;
    }
```
WDYT?

src/RelayProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+        requestConfig: RequestConfig
+    ): Promise<RelayingResult> {
+        const config = getEnvelopingConfig();
+
+        const { relayData, request } = envelopingRequest;
+
+        let callForwarderValue = await relayData.callForwarder;
+        let relayHubValue = await request.relayHub;
+        let onlyPreferredRelaysValue = requestConfig.onlyPreferredRelays;
+        let gasToSend = requestConfig.forceGasLimit;
```
franciscotobar, line 155:
```typescript
       const gasToSend = requestConfig.forceGasLimit ? BigNumber.from(requestConfig.forceGasLimit) : undefined;
```
And we need to change the import, instead of using BigNumberJs, we need to use the BigNumber from ethers.

src/RelayProvider.ts
```diff
@@ -0,0 +1,255 @@
+import type { Transaction } from 'ethers';
```
franciscotobar, line 1:
```suggestion
import { Transaction, BigNumber } from 'ethers';
```

src/RelayProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+import { JsonRpcProvider, TransactionReceipt } from "@ethersproject/providers";
+import type { ConnectionInfo } from "@ethersproject/web";
+import { getEnvelopingConfig } from './utils';
+import RelayClient, { RequestConfig } from './RelayClient';
+import log from 'loglevel';
+import { RelayHub__factory } from '@rsksmart/rif-relay-contracts';
+import AccountManager from './AccountManager';
+import type { UserDefinedEnvelopingRequest } from './common/relayRequest.types';
+import { useEnveloping } from './utils';
+import BigNumber from 'bignumber.js';
```
franciscotobar, line 12: Remove this. Since The BigNumber library needs to use here is BigNumber from ethers.

ironFe93: got it

src/RelayProvider.ts
ironFe93: got it

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+         * When using a RelayProvider, and the gas field is not manually set, then the original provider within calculates a ridiculously high gas.
+         * In order to avoid this, we created a new field called forceGas, which is optional.
+         * If the user wants to manually set the gas, then she can put a value in the forceGas field, otherwise it will be
+         * automatically estimated (correctly) by the RelayClient.
+         * The reason why we added this new forceGas field is because at this point, by reading the gas field,
+         * we cannot differentiate between a user-entered gas value from the auto-calculated (and a very poor calculation)
+         * value that comes from the original provider
+         */
+
+        const fullEnvelopingRequest: UserDefinedEnvelopingRequest = {
```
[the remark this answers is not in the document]
ironFe93: It is no longer called this way :)

src/RelayProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+                return this._ethSendTransaction(envelopingData.envelopingTx, requestConfig);
+            }
+
+            if (method === 'eth_getTransactionReceipt') {
+                return this.jsonRpcProvider.send(method, params);
+            }
+
+            if (method === 'eth_accounts') {
+                return this._getAccounts(method, params);
+            }
```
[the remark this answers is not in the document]
ironFe93: this is done, please have a look

src/RelayProvider.ts
ironFe93: True, I have removed this error and leaving it up to the inner provider.

src/RelayProvider.ts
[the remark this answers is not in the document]
antomor: @ironFe93 I think this is still valid, isn't it?

antomor (approved):
LGTM

PR #64, opened 2022-12-22
PP-598/RelayClient-relayTransaction

franciscotobar:
## What

- RelayClient - relayTransaction refactor
- RelayCllient - _calculateGasPrice refactor

## Why

- Project refactoring

## Refs

- [PP-598](https://rsklabs.atlassian.net/browse/PP-598)


test/RelayClient.test.ts
```diff
@@ -1176,5 +1129,403 @@ describe('RelayClient', function () {
...
+        } as unknown as typeof relayHubStub;
+        sandbox.stub(RelayHub__factory, 'connect').returns(relayHubStub);
+        const verifyWithRelayHub = relayClient._verifyWithRelayHub(relayWorkerAddress, FAKE_RELAY_TRANSACTION_REQUEST, fakeMaxPossibleGas);
+
+        expect(callStub).to.be.calledOnce;
+        await expect(verifyWithRelayHub).to.be.rejected;
+      });
+
+      it('should throw if enveloping deploy request will fail', async function () {
+        const callStub = sandbox.stub().throws(new Error());
```
franciscotobar, line 1515: In some testing I throw error without description or dummy description since there are multiple scenarios possibles that were covered on their specific test

franciscotobar:
I would like to ask your assistance to validate the naming for the private functions, I have like 3 functions with the prefix verify but Im not pretty sure about it.

src/RelayClient.ts
```diff
@@ -424,6 +420,147 @@ class RelayClient extends EnvelopingEventEmitter {
...
+      activeRelay = await selectNextRelay(httpClient);
+    }
+
+    throw Error(NOT_RELAYED_TRANSACTION);
+  }
+
+  private async _attemptRelayTransaction(
+    { relayInfo: { url }, hubInfo }: RelayInfo,
+    envelopingTx: EnvelopingTxRequest,
+    httpClient: HttpClient
```
franciscotobar, line 453: should I pass the client around, create a new instance whenever I need it, or create an instance during the constructor?

src/RelayClient.ts
```diff
@@ -232,68 +226,76 @@ class RelayClient extends EnvelopingEventEmitter {
...
       feesReceiver === constants.AddressZero ||
-      !isAddress(await feesReceiver)
+      !isAddress(feesReceiver)
     ) {
       throw new Error('FeesReceiver has to be a valid non-zero address');
     }
 
+    const currentTokenGas = BigNumber.from(tokenGas);
+
+    const newTokenGas = currentTokenGas.gt(constants.Zero) ?
```
franciscotobar, line 281: I think that we could add a forceTokenGas in the requestConfig

test/RelayClient.test.ts
jurajpiar: in that case you should cover each scenario, or pick one, I'd say, otherwise this may be a false negative.

test/utils.test.ts
```diff
@@ -117,10 +123,238 @@ describe('utils', function () {
...
+            recoverer: createRandomAddress(),
+            relayHub: '',
+            tokenAmount: 0,
+            tokenContract: '',
+            tokenGas: 0,
+            value: 0,
+            data: '',
+          },
+        },
+      };
```
jurajpiar, line 170: I'd suggest to use the existing fakes, or at least define fakes with "thruthly" values, with integers > 1 to make sure that calculations are performed correctly.

jurajpiar (approved):
Looks good

src/constants/errorMessages.ts
```diff
@@ -4,9 +4,11 @@ const MISSING_CONFIG_ROOT = `Could not read enveloping configuration. Make sure
 const MISSING_SMART_WALLET_ADDRESS =
   'Missing smart wallet address in requestConfig. Should be calculated before estimating the gas cost for a deploy transaction';
 const MISSING_CALL_FORWARDER = 'Missing call forwarder in EnvelopingRequest.';
+const NOT_RELAYED_TRANSACTION = 'Transaction was not relayed thorough any hub';
```
ironFe93, line 7:
```suggestion
const NOT_RELAYED_TRANSACTION = 'Transaction was not relayed through any hub';
```

src/utils.ts
```diff
@@ -24,19 +29,19 @@ const selectNextRelay = async (
...
 
-  for (const relayInfo of preferredRelays ?? []) {
+  for (const managerData of preferredRelays ?? []) {
     let hubInfo;
 
     try {
-      hubInfo = await httpClient.getChainInfo(relayInfo.url);
+      hubInfo = await httpClient.getChainInfo(managerData.url);
     } catch (error) {
       continue;
```
ironFe93, line 38: maybe we should print something instead of handling the error silently?

ironFe93 (approved):
lgtm but look at comments please

src/RelayClient.ts
ironFe93: I would create an instance on construction

test/RelayClient.test.ts
antomor: 👆🏽

src/RelayClient.ts
antomor: Yes, indeed. Without it, we cannot reproduce the exact same functionality we had before.

src/RelayClient.ts
antomor: IMHO I'd prefer to use the already created instance instead of creating a new instance for each method (especially if the method is a private one)

src/utils.ts
antomor: surely some logs may be useful here

test/utils.test.ts
```diff
@@ -117,10 +123,238 @@ describe('utils', function () {
...
+      } as unknown as RelayHub;
+      sinon.stub(RelayHub__factory, 'connect').returns(relayHubStub);
+    });
+
+    afterEach(function () {
+      sinon.restore();
+    });
+
+    it('Should perform checks on deploy transaction and not throw errors', function () {
+      // FIXME: this test sounds meaningless.
```
antomor, line 189: Indeed I don't find any difference between this one and the next one at line 209

test/utils.test.ts
```diff
@@ -117,10 +123,238 @@ describe('utils', function () {
...
+          transaction,
+          relayWorkerAddress
+        )
+      ).to.throw(
+        `Relay request Encoded data must be the same as Transaction data`
+      );
+    });
+
+    it('Should throw error if Relay Worker is not the same as Transaction sender', function () {
+      // FIXME: fix test
```
antomor, line 338: Is this comment still valid?

test/RelayClient.test.ts
```diff
@@ -240,7 +247,47 @@ describe('RelayClient', function () {
         expect(actualRelayRequestData).to.deep.equal(expectedRelayRequestData);
       });
 
+      it('should return tx request with tokenGas in request data if its gt than zero', async function () {
```
antomor, line 250:
```suggestion
      it("should return transaction request with 'tokenGas' in request data if it's greater than zero", async function () {
```

test/RelayClient.test.ts
```diff
@@ -240,7 +247,47 @@ describe('RelayClient', function () {
...
+        };
+
+        const {
+          relayRequest: { request: actualRelayRequest },
+        } = await relayClient._prepareHttpRequest(FAKE_HUB_INFO, FAKE_RELAY_REQUEST);
+
+        expect(actualRelayRequest).to.deep.equal(expectedRelayRequestData);
+      });
+
+      it('should return tx request with estimated tokenGas if its zero in request', async function () {
```
antomor, line 262:
```suggestion
      it("should estimate 'tokenGas' if it's zero in the request", async function () {
```

test/RelayClient.test.ts
```diff
@@ -621,61 +668,16 @@ describe('RelayClient', function () {
         );
       });
 
-      it('should return request with given tokenGas', async function () {
-        const {
-          request: { tokenGas: expectedTokenGas },
-        } = FAKE_DEPLOY_REQUEST;
+      //FIXME check if we need to have a different approach
```
antomor, line 671: Is this comment still valid? Which approach would you suggest and why do you find it wrong?

test/RelayClient.test.ts
```diff
@@ -1176,5 +1129,403 @@ describe('RelayClient', function () {
...
+      
+      it('should fail to relay transaction if details are missing', async function () {
+        relayClient._getEnvelopingRequestDetails = sandbox.stub().throws(new Error('missing details'));
+        const expectedTransaction = relayClient.relayTransaction(envelopingRequest, FAKE_REQUEST_CONFIG);
+
+        await expect(expectedTransaction).to.be.rejectedWith('missing details')
+      });
+
+      it('should fail to relay transaction if http request cannot be prepared', async function () {
+        relayClient._prepareHttpRequest = sandbox.stub().throws(new Error('http cannot be prepared'));
```
antomor, line 1191: I'd store the error message in a constant

test/RelayClient.test.ts
```diff
@@ -1176,5 +1129,403 @@ describe('RelayClient', function () {
...
+
+      it('should send transaction if cannot find it in pool or receipt', async function () {
+        relayClient._provider.getTransaction = sandbox.stub().returns(undefined);
+        relayClient._provider.getTransactionReceipt = sandbox.stub().returns(undefined);
+        await relayClient._broadcastTx(fakeSignedTx);
+
+        expect(sendTransactionStub.called).to.be.true;
+      });
+
+      it('should not send transaction if its in transaction pool', async function () {
```
antomor, line 1302:
```suggestion
      it("should not send transaction if the receipt is in the transaction pool", async function () {
```

test/RelayClient.test.ts
```diff
@@ -1176,5 +1129,403 @@ describe('RelayClient', function () {
...
+        const fakeResponse = {} as TransactionResponse;
+        relayClient._provider.getTransaction = sandbox.stub().returns(Promise.resolve(fakeResponse));
+        relayClient._provider.getTransactionReceipt = sandbox.stub().returns(undefined);
+        await relayClient._broadcastTx(fakeSignedTx);
+
+        expect(sendTransactionStub.called).to.be.false;
+      });
+
+      it('should not send transaction if receipt its found', async function () {
```
antomor, line 1311:
```suggestion
      it("should not send transaction if the receipt isn't found", async function () {
```

antomor (changes requested):
@franciscotobar thank you for the changes. I've left some minor comments plus some suggestions.

test/RelayClient.test.ts
franciscotobar: Ill cover each scenario

src/RelayClient.ts
franciscotobar: updated

test/RelayClient.test.ts
franciscotobar: updated

test/RelayClient.test.ts
franciscotobar: updated

test/RelayClient.test.ts
franciscotobar: this will be changed since Im taking on consideration all the scenarios now

test/RelayClient.test.ts
franciscotobar: the other approach was already validated, so its no longer needed to be fixed

test/RelayClient.test.ts
franciscotobar: updated

test/RelayClient.test.ts
franciscotobar: updated

test/utils.test.ts
franciscotobar: Im not pretty sure, @ironFe93, do you know if its still valid? Im also checking by myself.

test/utils.test.ts
franciscotobar: One is for deployment and the other one is regular transaction, should I keep both.

src/utils.ts
franciscotobar: updated

src/constants/errorMessages.ts
franciscotobar: updated

src/RelayClient.ts
franciscotobar: I updated, I added an extra param but Im not pretty sure about it, Im also thinking creating a new method for this, but I would like to read more opinions.

test/utils.test.ts
franciscotobar: updated

test/utils.test.ts
franciscotobar: I just check it and for me, its valid, it has a different worker address

test/RelayClient.test.ts
franciscotobar: updated

test/RelayClient.test.ts
franciscotobar: updated

src/RelayClient.ts
jurajpiar: Got no opinion on the `forceTokenGas` as I don't know the interactions between individual parts, but the `currentTokenGas.gt(constants.Zero)` can be written as `!currentTokenGas.isZero()` ;)

test/utils.test.ts
antomor: No, I wouldn't. I know that logically they're different scenarios, but if the request is exactly the same, so I wouldn't replicate it (we could add a comment to specify it).

test/utils.test.ts
antomor: so, in order to fix the test, should we change the worker address being used?

src/RelayClient.ts
antomor: I was referring to using the instance passed in the method, but since it's a class, this approach is fine also. We can consider changing it once we decide to go for a more functional approach 👍🏽

src/utils.ts
antomor: I'd log an error in case no hubInfo has been retrieved

src/RelayClient.ts
franciscotobar: updated but this is open to discussion

src/utils.ts
franciscotobar: updated

test/utils.test.ts
franciscotobar: updated

test/utils.test.ts
franciscotobar: but the test itself needs to throw an error what is currently doing, if I put the worker that is being used, it will not throw the error

src/RelayClient.ts
```diff
@@ -424,6 +418,158 @@ class RelayClient extends EnvelopingEventEmitter {
       estimatedGasCorrectionFactor
     );
   }
+
+  public async relayTransaction(
+    envelopingRequest: UserDefinedEnvelopingRequest,
+    requestConfig: RequestConfig
+  ): Promise<Transaction> {
```
ironFe93, line 425: Would it be possible to return TransactionResponse instead? https://docs.ethers.org/v5/api/providers/types/#providers-TransactionResponse

franciscotobar: its possible, based in our conversation but there is a chance that could return undefined so the typing will need to include it.

ironFe93: Actually, disregard this. I can wait for the trx receipt by using the provider anyway.

test/utils.test.ts
antomor: sure, got it!

src/RelayClient.ts
antomor: Sorry, my fault. At the beginning, I understood you were referring to `forceGas`. Indeed, we had to apply the logic about it, but now that I checked again, I'm sure we don't need `forceTokenGas` because that field will never be overwritten by the provider being a custom property.

antomor (approved):
LGTM

PR #68, opened 2023-01-05
PP-357/dapp-update

franciscotobar:
## What

- changes to support dApp refactoring

## Why

- required by dApp refactor

## Refs

- [update rif-relay-sample-dapp](https://rsklabs.atlassian.net/browse/PP-357)


src/RelayClient.ts
```diff
@@ -38,57 +44,83 @@ import { isDeployRequest } from './common/relayRequest.utils';
...
+      throw new Error('Field `from` is not defined in request body.');
+    }
+
+    if (!tokenContract) {
+      throw new Error('Field `tokenContract` is not defined in request body.');
+    }
+
+    if (!isDeployment) {
+      if (!relayData?.callForwarder) {
+        throw new Error(MISSING_CALL_FORWARDER);
```
antomor, line 98: Why do we store only this error message in the constant? I'd try to be consistent and do the same with the others.

src/RelayClient.ts
```diff
@@ -153,40 +168,35 @@ class RelayClient extends EnvelopingEventEmitter {
...
       callVerifier,
       feesReceiver: constants.AddressZero, // returns zero address and is to be completed when attempting to relay the transaction
       gasPrice,
     };
 
     const secondsNow = Math.round(Date.now() / 1000);
-    const validUntilTime = (
-      secondsNow + this._envelopingConfig.requestValidSeconds
-    );
+    const validUntilTime = request.validUntilTime ??
```
antomor, line 198: @franciscotobar thank you for this fix 👍🏽

package.json
```diff
@@ -54,11 +54,9 @@
     "package.json": "npx sort-package-json"
   },
   "dependencies": {
-    "@metamask/eth-sig-util": "^5.0.0",
     "@rsksmart/rif-relay-contracts": "github:rsksmart/rif-relay-contracts#feature/PP-278/PP-329/upgrade_contracts",
-    "axios": "~1.1.2",
+    "axios": "~0.27.2",
```
antomor, line 58: Should we add a task in the backlog to update the axios version?

src/RelayClient.ts
```diff
@@ -219,25 +229,25 @@ class RelayClient extends EnvelopingEventEmitter {
...
-    const accountManager = new AccountManager(this._provider, chainId);
+      request: { relayHub, tokenGas, nonce, gas, value },
+      relayData: { gasPrice },
+    } = envelopingRequest;
+
+    const provider = getProvider();
+    const accountManager = new AccountManager();
     const relayMaxNonce =
-      (await this._provider.getTransactionCount(callForwarder)) +
+      (await provider.getTransactionCount(relayWorkerAddress)) +
```
antomor, line 250: Another good fix, thank you 👍🏽

src/RelayClient.ts
```diff
@@ -286,160 +316,133 @@ class RelayClient extends EnvelopingEventEmitter {
...
-          log.debug('Relay Client - Relayed done');
+    log.debug('Relay Client - Estimating transaction');
+    log.debug(
+      `Relay Client - Relay Hub:${envelopingRequestDetails.request.relayHub.toString()}`
+    );
 
-          return transaction;
-        }
-      }
+    const activeRelay = await selectNextRelay(this._httpClient);
```
antomor, line 408: The FIXME needs to be applied here also

src/RelayClient.ts
```diff
@@ -286,160 +316,133 @@ class RelayClient extends EnvelopingEventEmitter {
...
 
-      activeRelay = await selectNextRelay(this._httpClient);
-    }
+    const envelopingTx = await this._prepareHttpRequest(
+      activeRelay!.hubInfo,
+      envelopingRequestDetails
+    );
```
antomor, line 413: Unless I'm missing something, it looks like this part is identical to what we have in the relayTransaction (L#365-376)

src/api/common/HttpWrapper.ts
```diff
@@ -50,21 +50,23 @@ const interceptors = {
...
       }
-      log.debug(`Configuration: ${stringify(config)}`);
+      logger().debug(`Configuration: ${stringify(config)}`);
 
       return Promise.reject(error);
     },
   },
 };
 
+const logger = () => log.getLogger('HttpWrapper');
```
antomor, line 68: I know it's the same, but I'd prefer to define the `logger` function at the top of the file.

src/utils.ts
```diff
@@ -1,41 +1,30 @@
...
 
-  for (const managerData of preferredRelays ?? []) {
+  for (const preferredRelay of preferredRelays ?? []) {
     let hubInfo;
+    let managerData;
 
     try {
-      hubInfo = await httpClient.getChainInfo(managerData.url);
+      hubInfo = await httpClient.getChainInfo(preferredRelay);
+      const relayHub = RelayHub__factory.connect(hubInfo.relayHubAddress, getProvider());
```
antomor, line 26: Why don't we check any more for the `ready` property in `hubInfo`

src/utils.ts
```diff
@@ -54,6 +43,123 @@ const selectNextRelay = async (
...
+    internalEstimationCorrection
+  );
+
+  return applyGasCorrectionFactor(
+    internalCallCost,
+    estimatedGasCorrectionFactor
+  );
+}
+
+const getSmartWalletAddress = async (
```
antomor, line 122: In the future, we should consider if we want to replace this mechanism with the one that doesn't use the factory to check if the result is the same and if we can avoid using the factory.

test/RelayClient.test.ts
```diff
@@ -515,6 +492,23 @@ describe('RelayClient', function () {
...
+            },
+          } as unknown as UserDefinedEnvelopingRequest
+        );
+
+        await expect(call).to.be.eventually.rejected;
+        await expect(call).to.be.eventually.rejectedWith(
+          'Field `index` is not defined in deploy body.'
+        );
+      });
+
```
antomor, line 511: I'd add a specific case if the index is 0

test/RelayClient.test.ts
```diff
@@ -189,12 +177,12 @@ describe('RelayClient', function () {
...
             relayHub: expectedRelayHubAddress,
           },
-        }, FAKE_REQUEST_CONFIG);
+        });
 
         expect(actualRelayHubAddress).to.equal(expectedRelayHubAddress);
       });
 
-      it('should return tx request with fees receiver in request data from given hub info if no fees receiver given in request', async function () {
+      it.skip('should return tx request with fees receiver in request data from given hub info if no fees receiver given in request', async function () {
```
antomor, line 185: How do we plan to fix the skipped test? Do we want to add a specific task for all of them?

antomor (changes requested):
@franciscotobar Thank you for these changes. I left some comments to keep in mind that there are still open points there.

package.json
franciscotobar: yes I think so, I tried but it was time consuming so I decided to follow the easy path.

src/RelayClient.ts
franciscotobar: updated

src/api/common/HttpWrapper.ts
franciscotobar: updated

src/utils.ts
franciscotobar: we do in line 33, but I did some refactor.

test/RelayClient.test.ts
franciscotobar: now I forgot to discuss it with you, maybe after the daily we can check it.

src/RelayClient.ts
franciscotobar: now is the same, but before it was different.

src/RelayClient.ts
franciscotobar: updated

src/RelayClient.ts
franciscotobar: updated

src/RelayClient.ts
franciscotobar: updated

src/utils.ts
franciscotobar: updated

src/utils.ts
franciscotobar: updated

test/RelayClient.test.ts
franciscotobar: updated

test/RelayClient.test.ts
```diff
@@ -865,70 +897,173 @@ describe('RelayClient', function () {
...
+
+      it('should build a transaction using a hub', async function () {
+        sandbox.stub(relayUtils, 'selectNextRelay').resolves(relayInfo);
+        relayClient._prepareHttpRequest = sandbox.stub().returns(FAKE_RELAY_TRANSACTION_REQUEST);
+        const enveloping = await relayClient._getHubEnvelopingTx(envelopingRequest);
+
+        expect(enveloping).to.be.deep.equal(expectedEnveloping);
+      });
+
+      it('should throw if there is no there is no hub available', async function () {
```
antomor, line 1052:
```suggestion
      it('should throw if there is no hub available', async function () {
```

antomor (approved):

antomor (approved):
LGTM

PR #69, opened 2023-01-18
PP-642/migrate-test-utils-server

franciscotobar:
## What

- Migrate utility files used by the server

## Why

- Necessary to integration test

## Refs

- [PP-642](https://rsklabs.atlassian.net/browse/PP-642)

### DIFF
--- a/src/AccountManager.ts
+++ b/src/AccountManager.ts
@@ -11,7 +11,17 @@ import {
 } from './typedRequestData.utils';
 
 export default class AccountManager {
+  private static instance: AccountManager;
+
   private _accounts: Wallet[] = [];
+
+  public static getInstance(): AccountManager {
+    if (!AccountManager.instance) {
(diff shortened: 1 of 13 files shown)


src/AccountManager.ts
```diff
@@ -11,8 +11,18 @@ import {
 } from './typedRequestData.utils';
 
 export default class AccountManager {
+  private static instance: AccountManager;
+
   private _accounts: Wallet[] = [];
 
+  public static getInstance(): AccountManager {
```
antomor, line 18: I think we should create a private constructor to avoid the creation of multiple instances

antomor (changes requested):
@franciscotobar We need to slightly change `accountManager` to implement the singleton

src/AccountManager.ts
franciscotobar: updated

antomor (approved):
LG

PR #72, opened 2023-02-06
Fix: update rif-relay-contracts dependency

antomor:
## What

- Fix relay contracts peer dependency

## Why

- The format was wrong (starting with `v2.0.0-beta.0` instead of `2.0.0-beta.0`)


franciscotobar (approved):
LGTM

PR #74, opened 2023-02-06
PP-633/relay-sever-tests

franciscotobar:
## What

- Token inclusion while doing gas estimation
- Access for calculateGasPrice

## Why

- Bug found during integration test, since value was added twice
- calculateGasPrice its just needed to use for relaying transactins

## Refs

- [PP-633](https://rsklabs.atlassian.net/browse/PP-633)


antomor (approved):
LGTM

PR #75, opened 2023-02-26
PP-645:replace axios

ironFe93:
## What

Replace axios with SuperAgent
## Why

Request library must be compatible on both client and server

### DIFF
--- a/src/api/common/HttpWrapper.ts
+++ b/src/api/common/HttpWrapper.ts
@@ -1,14 +1,13 @@
-import axios, {
-  AxiosError,
-  AxiosInstance,
-  AxiosRequestConfig,
-  AxiosResponse,
-} from 'axios';
+import SuperAgent, { Response, ResponseError } from 'superagent'
 import log, { LogLevelDesc } from 'loglevel';
 
-const logger = () => log.getLogger('HttpWrapper');
+const logger = log.getLogger('HttpWrapper');
 const LOGMAXLEN = 120;
 const DEFAULT_TIMEOUT = 30000;
+
+type HttpWrapperOpts = {
+  timeout?: number,
+}
 
 const getCircularReplacer = () => {
   const seen = new WeakSet();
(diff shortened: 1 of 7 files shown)


src/api/common/HttpWrapper.ts
```diff
@@ -29,78 +28,72 @@ const stringify = (something: unknown): string =>
...
-      logger().debug('relayTransaction response:', response);
-      const data = response.data as { error: string } | undefined;
-      if (data?.error) {
-        logger().error(`Error within response: ${data.error}`);
+    onErrorResponse: (response: Response) => {
+      logger.debug('relayTransaction response:', response.body);
+      const data = response.body as { error: string } | undefined;
+
+      if (!data) {
+        throw new Error('Got undefined response from relay');
```
antomor, line 36: Shouldn't we log this error here?

src/api/common/HttpWrapper.ts
```diff
@@ -29,78 +28,72 @@ const stringify = (something: unknown): string =>
...
+  private readonly _httpClient;
 
-    logger().setLevel(loglLevel);
+  private timeout;
 
-    this._httpClient.interceptors.response.use(
-      interceptors.logRequest.onResponse,
-      interceptors.logRequest.onError
-    );
+  constructor(opts: HttpWrapperOpts = {}, loglLevel: LogLevelDesc = 'error') {
```
antomor, line 68:
```suggestion
  constructor(opts: HttpWrapperOpts = {}, logLevel: LogLevelDesc = 'error') {
```

src/api/pricer/CoinBase.ts
```diff
@@ -38,17 +38,13 @@ export default class CoinBase extends BaseExchangeApi {
...
-      const axiosError = error as AxiosError;
-      const { response } = axiosError;
-      if (response) {
-        throw Error(
-          `CoinBase API status ${response.status}/${response.statusText}`
-        );
-      }
-      if (axiosError.request) {
-        throw Error('No response received from CoinBase API');
+      const { response } = error as ResponseError;
```
antomor, line 41: I'd log the error here, otherwise we won't be able to understand where it comes from, since we're generating a new error.

antomor (approved):
@ironFe93 Thank you for the changes. Apart from some minor comments that I left, LGTM

src/api/common/HttpWrapper.ts
ironFe93: done

src/api/common/HttpWrapper.ts
ironFe93: done

src/api/pricer/CoinBase.ts
ironFe93: I think the reason it was not logged to begin with is that the interceptor for the wrapper will log it on error. Let me know if I should still log this though.

PR #77, opened 2023-02-28
PP-708: fix http client

antomor:
## What

- Change the way we build the URL used for making the requests against the server

## Why

- Right now, the relative path of the URL isn't taken into account.


ironFe93 (approved):
LGTM

PR #79, opened 2023-03-10
PP-682/review error handler

ironFe93:
## What

check if an 'error' property is found

## Why

In order to throw the correct error

## Refs

### DIFF
--- a/src/api/common/HttpClient.ts
+++ b/src/api/common/HttpClient.ts
@@ -93,7 +93,7 @@ class HttpClient {
         url,
         this._stringifyEnvelopingTx(envelopingTx)
       );
-    log.info('relayTransaction response:', signedTx);;
+    log.info('relayTransaction response:', signedTx);
 
     if (error) {
       throw new Error(`Got error response from relay: ${error}`);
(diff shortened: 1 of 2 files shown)


src/api/common/HttpClient.ts
```diff
@@ -86,13 +88,19 @@ class HttpClient {
...
 
-    const { signedTx } =
+    const { signedTx, error, message } =
       await this._httpWrapper.sendPromise<SignedTransactionDetails>(
         url,
         this._stringifyEnvelopingTx(envelopingTx)
       );
     log.info('relayTransaction response:', signedTx);
 
+    const errMsg = error ?? message;
```
antomor, line 98: Since we're calling the `/relay` handler, the error should be stored in the `error` property, hence `message` should be always undefined. It looks like we do set the `message` property in the `/tokens` handler only, but we're using it from the client at all.

antomor (approved):
@ironFe93 Thank you for the in-depth analysis. I'm going to create a task to analyse if we can implement a proper error handler in the server also.
I've just left a small comment, but LGTM 👏🏽

src/api/common/HttpClient.ts
ironFe93: removed case for message.

PR #80, opened 2023-03-19
PP-635/relay-client-tests

franciscotobar:
## What

- Update the relay-client constructor to receive a http client
- Update the event registration and unregistration

## Why

- Part of the relay-client integration tests migration

## Refs

- [PP-635](https://rsklabs.atlassian.net/browse/PP-635)

### DIFF
--- a/src/RelayClient.ts
+++ b/src/RelayClient.ts
@@ -56,7 +56,11 @@ import {
   NOT_RELAYED_TRANSACTION,
 } from './constants/errorMessages';
 import EnvelopingEventEmitter, {
-  envelopingEvents,
+  EVENT_INIT,
+  EVENT_RELAYER_RESPONSE,
+  EVENT_SEND_TO_RELAYER,
+  EVENT_SIGN_REQUEST,
+  EVENT_VALIDATE_REQUEST,
 } from './events/EnvelopingEventEmitter';
 import { estimateRelayMaxPossibleGas } from './gasEstimator';
 import {
@@ -327,7 +331,7 @@ class RelayClient extends EnvelopingEventEmitter {
       metadata,
     };
 
-    this.emit(envelopingEvents['sign-request']);
+    this.emit(EVENT_SIGN_REQUEST);
     log.info(
       `Created HTTP ${
         isDeployment ? 'deploy' : 'relay'
@@ -378,7 +382,7 @@ class RelayClient extends EnvelopingEventEmitter {
       options?.signerWallet
     );
 
-    this.emit('init');
+    this.emit(EVENT_INIT);
     log.debug('Relay Client - Relaying transaction');
     log.debug(
       `Relay Client - Relay Hub:${envelopingTx.metadata.relayHubAddress.toString()}`
@@ -457,7 +461,7 @@ class RelayClient extends EnvelopingEventEmitter {
         managerData: { url },
         hubInfo,
       } = relayInfo;
-      this.emit('send-to-relayer');
+      this.emit(EVENT_SEND_TO_RELAYER);
       log.info(
         `attempting relay: ${JSON.stringify(
           relayInfo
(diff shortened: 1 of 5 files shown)


src/events/EnvelopingEventEmitter.ts
```diff
@@ -2,35 +2,32 @@ import EventEmitter from 'events';
...
-  enveloping: 'enveloping',
   'refresh-relays': 'refresh-relays', // TODO validate if its needed, since we do not refresh relays anymore
   'refreshed-relays': 'refreshed-relays', // TODO validate if its needed, since we do not refresh relays anymore
   'next-relay': 'next-relay',
   'sign-request': 'sign-request',
   'validate-request': 'validate-request',
   'send-to-relayer': 'send-to-relayer',
-  'relayer-response': 'relayer-response', // FIXME needs to have a boolean flag
+  'relayer-response': 'relayer-response',
 } as const;
```
antomor, line 12: What if we create constants/enum to store those events? E.g.:
```
const EVENT_INIT = 'init';
const EVENT_ENVELOPING = 'enveloping';
...
```
or:
```
e
enum Event {
  Init = "init",
  Enveloping = "enveloping",
  ...
}
```

src/events/EnvelopingEventEmitter.ts
```diff
@@ -2,35 +2,32 @@ import EventEmitter from 'events';
...
 type Event = keyof typeof events;
 
 class EnvelopingEventEmitter extends EventEmitter {
   registerEventListener(
-    eventType: Event,
-    eventHandle: (...args: unknown[]) => void
+    eventHandler: (event: Event, ...args: unknown[]) => void
   ) {
-    this.on(eventType, eventHandle);
+    this.on('enveloping', eventHandler);
```
antomor, line 20: Why are we hard-coding this event here?

src/RelayClient.ts
```diff
@@ -79,11 +79,11 @@ class RelayClient extends EnvelopingEventEmitter {
 
   private readonly _httpClient: HttpClient;
 
-  constructor() {
+  constructor(httpClient: HttpClient = new HttpClient()) {
     super();
 
     this._envelopingConfig = getEnvelopingConfig();
-    this._httpClient = new HttpClient();
+    this._httpClient = httpClient;
```
antomor, line 86: 👍🏼

antomor (changes requested):
@franciscotobar Thank you for the changes. I've just left doubts/questions.

src/events/EnvelopingEventEmitter.ts
franciscotobar: we can use that approach, I kept the same approach that was used before, but I dont see any issue by changing it.

src/events/EnvelopingEventEmitter.ts
franciscotobar: By using the event type, you will be just listening to that event when you register.  The current implementation allows you to listen to all of the envelopingEvents together, while in the previous implementation you were only listening to each event, I modified to keep the same behavior as before.

franciscotobar: the enveloping its no an event, its like the wrapper of the events.

src/events/EnvelopingEventEmitter.ts
franciscotobar: updated

src/events/EnvelopingEventEmitter.ts
franciscotobar: updated

antomor: > the enveloping its no an event, its like the wrapper of the events.

Yes, I was missing this 👍🏼

antomor (approved):
LGTM

src/events/EnvelopingEventEmitter.ts
antomor: couldn't find the related changes, so I updated the PR directly ;-)

PR #81, opened 2023-03-22
PP-753: Implement coingecko pricer

antomor:
## What

- (describe the changes)

## Why

- (the reason these changes are wanted)

## Refs

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rsksmart/rif-relay-client",
-  "version": "2.0.0-beta.4",
+  "version": "2.0.0-beta.5",
   "private": false,
   "description": "This project contains all the client code for the rif relay system.",
   "license": "MIT",


src/api/pricer/CoinGecko.ts
```diff
@@ -0,0 +1,65 @@
...
+  ) {
+    super('CoinGecko', CURRENCY_MAPPING, ['RIF', 'RBTC', 'tRif']);
+  }
+
+  async queryExchangeRate(
+    sourceCurrency: string,
+    targetCurrency: string
+  ): Promise<BigNumberJs> {
+    let response: CoinGeckoResponse;
+    // TODO: We could check if the target currency is supported by calling the /simple/supported_vs_currencies API
```
antomor, line 33: I didn't implement it for 2 main reasons:
- right now we use it with `usd` only
- I didn't want to make an additional call if not strictly necessary, since there is a rate limit for the free plan

src/api/pricer/CoinBase.ts
```diff
@@ -14,15 +14,14 @@ export type CoinBaseResponse = {
...
-  RBTC: 'RBTC',
   TRIF: 'RIF',
 };
 
 export default class CoinBase extends BaseExchangeApi {
   private readonly _httpWrapper: HttpWrapper;
 
   constructor(httpWrapper: HttpWrapper = new HttpWrapper()) {
-    super('CoinBase', CURRENCY_MAPPING, ['RIF', 'RBTC', 'tRif']);
+    super('CoinBase', CURRENCY_MAPPING, ['RIF', 'tRif']);
```
franciscotobar, line 24: why did we remove it from here?  should we able to recover from the error and test the next api?

franciscotobar (approved):
LGTM, my only concern is about the recovery from errors.

src/api/pricer/CoinBase.ts
antomor: Yes, you're right, but I just wanted to fix the issue ASAP. I'll create a task to implement some sort of selection strategy, since there are some APIs (coinbase and coingecko) that support the same token (RIF)

antomor: [Task](https://rsklabs.atlassian.net/browse/PP-755) created

PR #82, opened 2023-03-24
PP-754: make the readonly properties public to allow developers to use it

antomor:
## What

- Change the relayClient to expose the httpClient
- Change the httpClient to expose the httpWrapper
- Change the httpWrapper to expose the superAgent instance

## Why

- It would allow developers to customize the requests as they want (e.g.: set request headers)


jurajpiar (approved):
LG

PR #84, opened 2023-04-05
PP644: implement relay server selector

AndresQuijano:
## What

- Add flexibility for Dapp developers to manage the way the select the relay by introducing two changes:
1. Create a private property and modify the constructor of the RelayClient (the class) in order to receive an optional parameter with the string representation of the relay URL.
2. Create a generator that is externally exposed so any Dapp developer can call next() and get an instance of the RelayClient with the next relay already set.

## Why

- Add flexibility without introducing breaking changes.

##Note
This change doesn't implement the logic of prioritization of relays (deciding which is the more convenient relay to select next based on criteria like the success/failure rate)

## Refs

- [PP-651](https://rsklabs.atlassian.net/browse/PP-651)

### DIFF
--- a/src/RelayClient.ts
+++ b/src/RelayClient.ts
@@ -67,7 +67,7 @@ import {
   estimateInternalCallGas,
   estimateTokenTransferGas,
   getSmartWalletAddress,
-  selectNextRelay,
+  // selectNextRelay,
   validateRelayResponse,
 } from './utils';
 
@@ -83,7 +83,10 @@ class RelayClient extends EnvelopingEventEmitter {
 
   private readonly _httpClient: HttpClient;
 
-  constructor(httpClient: HttpClient = new HttpClient()) {
+  constructor(
+    httpClient: HttpClient = new HttpClient(),
+    private _relayServerUrl?: string
+  ) {
     super();
 
     this._envelopingConfig = getEnvelopingConfig();
@@ -150,12 +153,14 @@ class RelayClient extends EnvelopingEventEmitter {
       throw new Error('No call verifier present. Check your configuration.');
     }
 
-    const relayServerUrl = this._envelopingConfig.preferredRelays.at(0);
+    if (!this._relayServerUrl) {
+      const relayServerUrl = this._envelopingConfig.preferredRelays.at(0);
 
-    if (!relayServerUrl) {
-      throw new Error(
-        'Check that your configuration contains at least one preferred relay with url.'
-      );
+      if (!relayServerUrl) {
+        throw new Error(
+          'Check that your configuration contains at least one preferred relay with url.'
+        );
+      }
     }
 
     const gasPrice: PromiseOrValue<BigNumberish> =
@@ -442,7 +447,7 @@ class RelayClient extends EnvelopingEventEmitter {
     );
 
     //FIXME we should implement the relay selection strategy
-    const activeRelay = await selectNextRelay(this._httpClient);
+    const activeRelay = await this._getRelayServer(this._httpClient);
 
     const envelopingTx = await this._prepareHttpRequest(
       activeRelay.hubInfo,
@@ -454,6 +459,55 @@ class RelayClient extends EnvelopingEventEmitter {
       activeRelay,
       envelopingTx,
     };
+  }
+
+  private async _getRelayServer(httpClient: HttpClient): Promise<RelayInfo> {
+    if (this._relayServerUrl) {
+      return await this._getRelayInfo(httpClient, this._relayServerUrl);
+    }
+
+    const { preferredRelays } = getEnvelopingConfig();
+
+    for (const preferredRelay of preferredRelays ?? []) {
+      try {
+        return await this._getRelayInfo(httpClient, preferredRelay);
+      } catch (error) {
+        log.warn('Failed to getChainInfo from hub', error);
+        continue;
+      }
+    }
+
+    log.error('No more hubs available to select');
+
+    throw new Error('No more hubs available to select');
+  }
+
+  private async _getRelayInfo(httpClient: HttpClient, relayServerUrl: string) {
+    try {
+      const hubInfo = await httpClient.getChainInfo(relayServerUrl);
+
+      if (hubInfo.ready) {
+        const relayHub = RelayHub__factory.connect(
+          hubInfo.relayHubAddress,
+          getProvider()
+        );
+
+        const managerData = await relayHub.getRelayInfo(
+          hubInfo.relayManagerAddress
+        );
+
+        return {
+          hubInfo,
+          managerData,
+        };
+      }
+
+      throw new Error(`Hub is not ready`); //TODO validate this error description
+    } catch (error: unknown) {
+      throw new Error(
+        `Failed to getChainInfo from hub ${(error as Error).message}`
+      );
+    }
   }
 
   private async _attemptRelayTransaction(
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -27,10 +27,29 @@ import {
   MISSING_SMART_WALLET_ADDRESS,
   MISSING_CALL_FORWARDER,
 } from './constants/errorMessages';
+import RelayClient from './RelayClient';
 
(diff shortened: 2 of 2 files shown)


src/RelayClient.ts
```diff
@@ -456,40 +452,95 @@ class RelayClient extends EnvelopingEventEmitter {
...
+      }
+    }
+
+    log.error('No more hubs available to select');
+
+    throw new Error('No more hubs available to select');
+  }
+
+  private async _getRelayInfo(relayServerUrl = this._relayServerUrl) {
+    try {
```
franciscotobar, line 477: I don't see any benefit of having this try/catch here.

src/RelayClient.ts
```diff
@@ -150,12 +155,14 @@ class RelayClient extends EnvelopingEventEmitter {
...
 
-    const relayServerUrl = this._envelopingConfig.preferredRelays.at(0);
+    if (!this._relayServerUrl) {
+      const relayServerUrl = this._envelopingConfig.preferredRelays.at(0);
 
-    if (!relayServerUrl) {
-      throw new Error(
-        'Check that your configuration contains at least one preferred relay with url.'
-      );
+      if (!relayServerUrl) {
```
franciscotobar, line 161:
```suggestion
      if (!this._envelopingConfig.preferredRelays.length) {
```
And will remove the variable.

src/RelayClient.ts
```diff
@@ -53,7 +53,6 @@ import type { EnvelopingTxRequest } from './common/relayTransaction.types';
 import {
   MISSING_CALL_FORWARDER,
   MISSING_REQUEST_FIELD,
-  NOT_RELAYED_TRANSACTION,
```
franciscotobar, line 56: if we are no longer using this error, we should get rid of it.

src/RelayClient.ts
```diff
@@ -442,7 +438,7 @@ class RelayClient extends EnvelopingEventEmitter {
     );
 
     //FIXME we should implement the relay selection strategy
```
franciscotobar, line 440: is this fixme still valid?

src/RelayClient.ts
```diff
@@ -456,40 +452,95 @@ class RelayClient extends EnvelopingEventEmitter {
     };
   }
 
+  private async _getRelayServer(): Promise<RelayInfo> {
+    if (this._relayServerUrl) {
+      return await this._getRelayInfo();
```
franciscotobar, line 457:
```suggestion
      return await this._getRelayInfo(this._relayServerUrl);
```

src/RelayClient.ts
```diff
@@ -456,40 +452,95 @@ class RelayClient extends EnvelopingEventEmitter {
...
+        continue;
+      }
+    }
+
+    log.error('No more hubs available to select');
+
+    throw new Error('No more hubs available to select');
+  }
+
+  private async _getRelayInfo(relayServerUrl = this._relayServerUrl) {
```
franciscotobar, line 476:
```suggestion
  private async _getRelayInfo(relayServerUrl: string) {
```

src/RelayClient.ts
```diff
@@ -456,40 +452,95 @@ class RelayClient extends EnvelopingEventEmitter {
...
+    try {
+      const hubInfo = await this._httpClient.getChainInfo(
+        relayServerUrl as string
```
franciscotobar, line 479:
```suggestion
        relayServerUrl
```

src/RelayClient.ts
```diff
@@ -83,13 +81,20 @@ class RelayClient extends EnvelopingEventEmitter {
...
     super();
 
     this._envelopingConfig = getEnvelopingConfig();
     this._httpClient = httpClient;
   }
 
+  public getRelayServerUrl() {
+    return this._relayServerUrl;
+  }
+
```
antomor, line 97: I guess that if`_relayServerUrl` isn't provided at construction time, this is undefined, isn't it?

src/RelayClient.ts
```diff
@@ -83,13 +81,20 @@ class RelayClient extends EnvelopingEventEmitter {
 
   private readonly _httpClient: HttpClient;
 
-  constructor(httpClient: HttpClient = new HttpClient()) {
+  constructor(
+    httpClient: HttpClient = new HttpClient(),
+    private _relayServerUrl?: string
```
antomor, line 86: What's the purpose of having the additional property here? I'm trying to track the property usage, but IMHO it would be the same if the client was configured to with a single server only.

src/utils.ts
```diff
@@ -21,48 +20,35 @@ import type {
...
+    index++;
+    const relayClient = new RelayClient(httpClient, url);
 
-  throw new Error('No more hubs available to select');
+    if (index < preferredRelays.length) {
+      yield { relayClient, status: `OK` };
+    } else {
+      index = 0;
+      yield { relayClient, status: `End of list of preferred relays` };
+    }
```
antomor, line 50: If we want to return a status, I think we should use constants or enum to store them

antomor: Furthermore, it's not clear to me what's the purpose of the status. I think the end of the returned values already explains what the status includes, doesn't it?

test/utils.test.ts
```diff
@@ -650,4 +583,63 @@ describe('utils', function () {
...
+      for (let i = 0; i < preferredRelays.length; i++) {
+        relayGenerator.next();
+      }
+
+      const next = relayGenerator.next();
+
+      expect(next.value.relayClient.getRelayServerUrl()).equal(
+        preferredRelays[0]
+      );
+    });
```
antomor, line 643: Should we add a test where we use the generator in a for, without explicitly calling `next`?

src/RelayClient.ts
AndresQuijano: Right, changed.

src/RelayClient.ts
AndresQuijano: Changed.

src/RelayClient.ts
AndresQuijano: Done.

src/RelayClient.ts
AndresQuijano: I think so, because we haven't implemented a real strategy here. We are only iterating the array of preferred relays.

src/RelayClient.ts
AndresQuijano: Done.

src/RelayClient.ts
AndresQuijano: Done.

src/RelayClient.ts
AndresQuijano: Done.

src/RelayClient.ts
AndresQuijano: Yes, it is.

src/RelayClient.ts
```diff
@@ -150,9 +155,10 @@ class RelayClient extends EnvelopingEventEmitter {
       throw new Error('No call verifier present. Check your configuration.');
     }
 
-    const relayServerUrl = this._envelopingConfig.preferredRelays.at(0);
-
-    if (!relayServerUrl) {
+    if (
+      !this._relayServerUrl &&
+      !this._envelopingConfig.preferredRelays.at(0)
```
franciscotobar, line 160:
```suggestion
     !this._envelopingConfig.preferredRelays.length
```

test/utils.test.ts
```diff
@@ -650,4 +583,38 @@ describe('utils', function () {
       expect(result).to.be.true;
     });
   });
+
+  describe('getRelayClientGenerator', function () {
```
franciscotobar, line 587: We should include a test where it fails, for example that the list is undefined. Or maybe we should handle that error in the generator.

test/utils.test.ts
antomor: @AndresQuijano I think this is still valid

AndresQuijano: Now I included a test without calling next().

antomor (approved):
LGTM

PR #85, opened 2023-04-05
PP-755/pricer-api-selection

franciscotobar:
## What

- Multiple apis to query for a token price
- Since there is no state in the pricer, the class was replaced by functions.
- Priority queue was implemented

## Why

- Right now RIF Token is supported by both Coinbase and CoinGecko clients, but the first one is selected only, and if it fails, we don’t get the price at all. We should have a strategy that implies the usage of the additional clients if available.

## Refs

- [PP-755](https://rsklabs.atlassian.net/browse/PP-755)


test/pricer/pricer.test.ts
```diff
@@ -0,0 +1,152 @@
...
+        fakeRbtcUsd
+      );
+      exchangeApiFactoryStub.withArgs('coinCodex').returns(coinCodexStub);
+      coinGeckoStub.queryExchangeRate
+        .withArgs(RBTC_SYMBOL, INTERMEDIATE_CURRENCY)
+        .rejects();
+
+      const beforeTokenToApi = tokenToApi[RBTC_SYMBOL];
+      const exchangeRate = await getExchangeRate(RIF_SYMBOL, RBTC_SYMBOL);
+      const afterTokenToApi = tokenToApi[RBTC_SYMBOL];
```
antomor, line 116: NIT: although it works, I think that we could have tested the same scenario by checking if the subsequent call is performed against the same pricer, instead of checking for an implementation detail

src/api/pricer/CoinBase.ts
```diff
@@ -12,29 +13,31 @@ export type CoinBaseResponse = {
...
-
 export default class CoinBase extends BaseExchangeApi {
-  private readonly _httpWrapper: HttpWrapper;
+  private _url: URL;
 
-  constructor(httpWrapper: HttpWrapper = new HttpWrapper()) {
-    super('CoinBase', CURRENCY_MAPPING, ['RIF', 'tRif']);
-    this._httpWrapper = httpWrapper;
+  constructor(
+    private readonly _httpWrapper: HttpWrapper = new HttpWrapper(),
```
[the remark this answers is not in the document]
franciscotobar: Ill make it public so the headers can be changed.

src/api/pricer/CoinBase.ts
```diff
@@ -12,29 +13,31 @@ export type CoinBaseResponse = {
...
     sourceCurrency: string,
     targetCurrency: string
   ): Promise<BigNumberJs> {
-    const upperCaseTargetCurrency = targetCurrency.toUpperCase();
     let response: CoinBaseResponse;
+    const sourceCurrencyName = this._getCurrencyName(sourceCurrency);
+    const targetCurrencyName = targetCurrency.toUpperCase();
 
     try {
+      const url = new URL(this._priceApiPath, this._url);
```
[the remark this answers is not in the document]
franciscotobar: The only reason that I put in the constructor was to validate it from the moment that the class is created, if not, we will just realize that the url is bad until we try a request.

src/pricer/ExchangeApiFactory.ts
```diff
@@ -0,0 +1,27 @@
...
+  CoinBase,
+  CoinCodex,
+  CoinGecko,
+  RdocExchange,
+  TestExchange,
+} from '../api';
+import type { ExchangeApiName } from './utils';
+
+class ExchangeApiFactory {
+  static getExchangeApi(exchageApiType: ExchangeApiName) {
```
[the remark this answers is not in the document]
franciscotobar: updated

src/pricer/ExchangeApiFactory.ts
```diff
@@ -0,0 +1,27 @@
...
+        return new CoinGecko();
+      case 'coinCodex':
+        return new CoinCodex();
+      case 'rdocExchange':
+        return new RdocExchange();
+      case 'testExchange':
+        return new TestExchange();
+    }
+  }
+}
```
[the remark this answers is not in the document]
franciscotobar: I did not include a default value since typescript is the one that is validating that.

test/pricer/pricer.test.ts
franciscotobar: cool, Ill do that. thanks

franciscotobar: updated

src/pricer/ExchangeApiFactory.ts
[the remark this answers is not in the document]
franciscotobar: updated

src/api/pricer/CoinCodex.ts
```diff
@@ -0,0 +1,58 @@
...
+      if (!response) {
+        throw new Error('No response received from CoinCodex API');
+      }
+
+      throw Error(`CoinCodex API status ${response.status}`);
+    }
+
+    const { last_price_usd: lastPriceUsd }: CoinCodexResponse = response;
+
+    if (targetCurrencyName !== 'USD') {
```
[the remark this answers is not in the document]
franciscotobar: I dont understand this.

src/api/pricer/CoinCodex.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export type CoinCodexResponse = {
+  last_price_usd: string;
+};
+
+export default class CoinCodex extends BaseExchangeApi {
+  private _url: URL;
+
+  constructor(
+    private readonly _httpWrapper: HttpWrapper = new HttpWrapper(),
```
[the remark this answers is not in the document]
franciscotobar: updated

franciscotobar:
I think that we need to discuss about the way of handling URL's in the Api's.

src/api/pricer/CoinCodex.ts
```diff
@@ -0,0 +1,58 @@
...
+      throw Error(`CoinCodex API status ${response.status}`);
+    }
+
+    const { last_price_usd: lastPriceUsd }: CoinCodexResponse = response;
+
+    if (targetCurrencyName !== 'USD') {
+      throw Error(
+        `Exchange rate for currency pair ${sourceCurrency}/${targetCurrency} is not available`
+      );
+    }
```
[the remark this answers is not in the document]
antomor: And what if `last_price_usd` is undefined? should we check for it also before using its value to create a BigNumberJs?

src/api/pricer/CoinCodex.ts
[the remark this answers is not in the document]
antomor: Sorry if it's not clear, It's related to the previous comment. Should we check if `lastPriceUsd` is defined before using it to create a BigNumberJs?

src/pricer/ExchangeApiFactory.ts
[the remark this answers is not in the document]
antomor: thank you, I see that you opted for the complete version with constructor args 🙏🏼

src/pricer/pricer.ts
```diff
@@ -0,0 +1,86 @@
...
+
+        return exchangeRate;
+      }
+    } catch (e: unknown) {
+      log.error(e);
+    }
+  }
+
+  return BigNumberJs(0);
+};
```
antomor, line 84: I'd simplify this a bit by removing the try/catch
```suggestion
  for (const api of exchangesApi) {
    const exchangeApi = createExchangeApi(api);


      const exchangeRate = await exchangeApi.queryExchangeRate(
        sourceCurrency,
        targetCurrency
      ).catch((e: unknown) => {
        log.error(e);
        
        return undefined;
      });

      if (exchangeRate) {
        tokenToApi[sourceCurrency.toUpperCase()] = [
          ...new Set([api, ...exchangesApi]),
        ];

        return exchangeRate;
      }
      
      log.info(`No exchange rate found for ${sourceCurrency}/${targetCurrency} using API ${api}`); // or log.debug, I'm not sure 😅
  }

  return BigNumberJs(0);
};
```

src/pricer/utils.ts
antomor, line 1: 🚀

antomor:
@franciscotobar 
> I think that we need to discuss about the way of handling URL's in the Api's.

I think that it's a good strategy indeed, so in this case, I'd:
1. avoid the creation of the new object if not required (e.g.: see `CoinCodex`)
2. create the new URL just with `const urlWithSearchParams = new URL(this._url)` (no need to specify both constructor args) if needed, so when we need to add any additional property to the previously created URL

src/api/pricer/CoinCodex.ts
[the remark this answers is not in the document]
franciscotobar: Well that was the way that I have it before but after thinking about it, at this point we already validate that CoinCodex replied to us, and unless that they change the API, the last_price_usd will be there, I could move again the validation prior the return.

src/pricer/pricer.ts
franciscotobar: I think that here we could include a notifier as well for the grafana logs, so we could know if an API failed and validate if something needs to be changed to make it work again.

src/api/pricer/CoinGecko.ts
```diff
@@ -51,14 +52,16 @@ export default class CoinGecko extends BaseExchangeApi {
...
-        `Exchange rate for currency pair ${sourceCurrency}/${currencyId} is not available`
+        `Exchange rate for currency pair ${sourceCurrency}/${targetCurrency} is not available`
       );
     }
 
     // It cannot be undefined due to the previous check
-    const conversionRate = response[sourceCurrency]?.[currencyId] as string;
+    const conversionRate = response[sourceCurrencyName]?.[
+      targetCurrencyName
+    ] as string;
```
jurajpiar, line 64: I know it's not yours, just to make it cleaner, you could swap these and remove the comment.
```suggestion
const conversionRate = response[sourceCurrencyName]?.[
  targetCurrencyName
];

if (!conversionRate) {
  throw Error(
    `Exchange rate for currency pair ${sourceCurrency}/${targetCurrency} is not available`
  );
}
```

or, since we already use the base class (pretty much) only for enforcing error (in _getCurrencyName), we could use something like:
```suggestion
return this._validRate(response[sourceCurrencyName]?.[
  targetCurrencyName
] as string, sourceCurrency, targetCurrency);
```
and in base class
```typescript
protected _validRate(rate: string, sourceCurrency: string, targetCurrency: string): BigNumberJs {
    if (!rate) {
      throw Error(
        `Exchange rate for currency pair ${sourceCurrency}/${targetCurrency} is not available`
      );
    }
    
    return BigNumberJs(rate);
  }
```

src/api/pricer/CoinGecko.ts
```diff
@@ -51,14 +52,16 @@ export default class CoinGecko extends BaseExchangeApi {
...
 
     return BigNumberJs(conversionRate);
```
jurajpiar, line 66: I'd like to point out that we're using bigNumberJs here even tho no calculations are yet being performed, so perhaps we should use BigNumber as return type.

src/pricer/pricer.ts
```diff
@@ -0,0 +1,89 @@
...
+    intermediateCurrency
+  );
+
+  const targetExchangeRate = await queryExchangeRate(
+    targetCurrency,
+    intermediateCurrency
+  );
+
+  if (sourceExchangeRate.isEqualTo(0) || targetExchangeRate.isEqualTo(0)) {
+    const message = `Currency conversion for pair ${sourceCurrency}:${targetCurrency} not found in current exchange api`;
```
jurajpiar, line 28: this kinda duplicates the error in each api class. I know that the check is different, but I'd say it should be handled in a single place only.

src/pricer/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+builders.set(
+  'coinGecko',
+  (args?: CoinCodexConstructorArgs) => new CoinCodex(...(args || []))
+);
+builders.set(
+  'coinCodex',
+  (args?: CoinGeckoConstructorArgs) => new CoinGecko(...(args || []))
+);
+builders.set('rdocExchange', () => new RdocExchange());
+builders.set('testExchange', () => new TestExchange());
```
jurajpiar, line 52:
```suggestion
const apiHandlers: Record<ExchangeApiName, (args?: ConstructorArgs) => BaseExchangeApi> = {
  coinGecko: (args: CoinGeckoConstructorArgs = []) => new CoinGecko(...args),
  coinCodex: (args: CoinCodexConstructorArgs = []) => new CoinGecko(...args),
  coinBase: (args: CoinBaseConstructorArgs = []) => new CoinGecko(...args),
  rdocExchange: () => new RdocExchange(),
  testExchange: () => new TestExchange()
}
```

src/pricer/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+
+function createExchangeApi(
+  exchangeApiType: ExchangeApiName,
+  args?: ConstructorArgs
+): BaseExchangeApi {
+  const fn = builders.get(exchangeApiType);
+  if (fn) return fn(args);
+
+  throw new Error(`${exchangeApiType} is not mapped`);
```
jurajpiar, line 61: not a biggie, but prob cleaner as
```suggestion
  if (!fn) throw new Error(`${exchangeApiType} is not mapped`);
  
  return fn(args);
```
to keep the return statement to the end

src/pricer/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+  (args?: CoinCodexConstructorArgs) => new CoinCodex(...(args || []))
+);
+builders.set(
+  'coinCodex',
+  (args?: CoinGeckoConstructorArgs) => new CoinGecko(...(args || []))
+);
+builders.set('rdocExchange', () => new RdocExchange());
+builders.set('testExchange', () => new TestExchange());
+
+function createExchangeApi(
```
jurajpiar, line 54: actually I'd consider value of this function as it only adds the error on non-existent api, but the only place this is used is in the pricer, which will then error if any one of them does not initialise (guessing that's the idea for the error handling anyway), so will need custom handling anyway. I'd just get it [where it is required](https://github.com/rsksmart/rif-relay-client/pull/85/files#diff-38cd6bf9d92f33682673b200915d827e7231f2bdbc538271a4c719e6cba37bf0R63) directly as `exchangeApis[exchangeApiType]()`

src/api/pricer/CoinGecko.ts
franciscotobar: what are you referring? BigNumberJs or BigNumber from ethers, as return type? BigNumber from ethers does not handle decimals.

src/pricer/utils.ts
franciscotobar: updated

src/pricer/utils.ts
franciscotobar: updated

franciscotobar: updated

src/pricer/pricer.ts
franciscotobar: well I just checked and the errors are different, are similar but describe different issues on different points, and I think that cannot be merged or centralized.

src/api/pricer/CoinGecko.ts
jurajpiar: ah of course, you are right.

test/pricer/pricer.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    function stubExchangeRate(
+      exchangeApi: SinonStubbedInstance<BaseExchangeApi>,
+      sourceCurrency: string,
+      targetCurrency: string,
+      exchangeRate: BigNumberJs
+    ) {
+      exchangeApi.queryExchangeRate
+        .withArgs(sourceCurrency, targetCurrency)
+        .resolves(exchangeRate);
+    }
```
jurajpiar, line 70: I wouldn't do this. It doesn't do anything special and it just executes one fn with the args we're passing to it.

src/pricer/pricer.ts
```diff
@@ -0,0 +1,93 @@
...
+const queryExchangeApis = async (
+  exchangesApi: ExchangeApiName[],
+  sourceCurrency: string,
+  targetCurrency: string
+): Promise<BigNumberJs> => {
+  for (const api of exchangesApi) {
+    const exchangeApi = createExchangeApi(api);
+
+    const exchangeRate = await exchangeApi
+      .queryExchangeRate(sourceCurrency, targetCurrency)
```
jurajpiar, line 70: this will throw without catching if no api is found using `createExchangeApi`, which is what I meant in https://github.com/rsksmart/rif-relay-client/pull/85/files#r1163974138

src/api/pricer/CoinBase.ts
```diff
@@ -34,10 +34,8 @@ export default class CoinBase extends BaseExchangeApi {
     const targetCurrencyName = targetCurrency.toUpperCase();
 
     try {
-      const url = new URL(this._priceApiPath, this._url);
-      url.searchParams.append('currency', sourceCurrencyName);
       response = await this._httpWrapper.sendPromise<CoinBaseResponse>(
-        url.toString()
+        `${this._url.toString()}?currency=${sourceCurrencyName}`
```
antomor, line 38: I think that in this case it was right to create a new one, since we adding a new search param there

src/api/pricer/CoinGecko.ts
```diff
@@ -39,11 +39,8 @@ export default class CoinGecko extends BaseExchangeApi {
     const targetCurrencyName = targetCurrency.toLowerCase();
 
     try {
-      const url = new URL(this._priceApiPath, this._url);
-      url.searchParams.append('ids', sourceCurrencyName);
-      url.searchParams.append('vs_currencies', targetCurrencyName);
       response = await this._httpWrapper.sendPromise<CoinGeckoResponse>(
-        url.toString()
+        `${this._url.toString()}?ids=${sourceCurrencyName}&vs_currencies=${targetCurrencyName}`
```
antomor, line 43: Same here, if we need to add new search params, let's create a new URL

antomor (changes requested):

src/pricer/pricer.ts
```diff
@@ -0,0 +1,93 @@
...
+    throw Error(`Token ${sourceCurrency} does not have an API to query`);
+  }
+
+  log.debug(`${exchangeApis.toString()} found for token ${sourceCurrency}`);
+
+  return queryExchangeApis(exchangeApis, sourceCurrency, targetCurrency);
+};
+
+const queryExchangeApis = async (
+  exchangesApi: ExchangeApiName[],
```
jurajpiar, line 62:
```suggestion
  exchangeApis: ExchangeApiName[],
```

test/pricer/pricer.test.ts
franciscotobar: updated

src/api/pricer/CoinBase.ts
franciscotobar: updated

src/api/pricer/CoinGecko.ts
franciscotobar: updated

src/pricer/pricer.ts
franciscotobar: updated

src/pricer/pricer.ts
antomor: @franciscotobar I think this is a good idea, let's change the format to be similar to the [one used in the server](https://github.com/rsksmart/rif-relay-server/blob/320b81165f21020b81b712e817833f69dc48743b/src/ReplenishFunction.ts#L9-L10).
After doing it, we should get @david-iov involved to create the related notification.

src/api/pricer/CoinCodex.ts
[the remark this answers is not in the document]
antomor: this is a duplication of [this](https://github.com/rsksmart/rif-relay-client/pull/85#discussion_r1160572068)

jurajpiar (approved):
LG

antomor (approved):

PR #86, opened 2023-05-03
fix/verify_with_relay_hub

franciscotobar:
## What

- Update the static client validation with the relayHub

## Why

- The server tries increasing the gas limit with an increase factor and the client does not.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,4 +1,11 @@
-import { BigNumberish, BigNumber, Transaction, constants, utils } from 'ethers';
+import {
+  BigNumberish,
+  BigNumber,
+  Transaction,
+  constants,
+  utils,
+  PopulatedTransaction,
+} from 'ethers';
 import { BigNumber as BigNumberJs } from 'bignumber.js';
 import {
   ICustomSmartWalletFactory__factory,
@@ -24,6 +31,7 @@ import type {
 import {
   MISSING_SMART_WALLET_ADDRESS,
   MISSING_CALL_FORWARDER,
+  GAS_LIMIT_EXCEEDED,
 } from './constants/errorMessages';
 import RelayClient from './RelayClient';
 import type { HttpClient } from './api/common';
@@ -31,6 +39,8 @@ import type { HttpClient } from './api/common';
 const INTERNAL_TRANSACTION_ESTIMATED_CORRECTION = 20000; // When estimating the gas an internal call is going to spend, we need to substract some gas inherent to send the parameters to the blockchain
 const ESTIMATED_GAS_CORRECTION_FACTOR = 1;
 const SHA3_NULL_S = utils.keccak256('0x');
+const FACTOR = 0.25;
+const GAS_VERIFICATION_ATTEMPTS = 5;
 
 const getRelayClientGenerator = function* (httpClient?: HttpClient) {
   const { preferredRelays } = getEnvelopingConfig();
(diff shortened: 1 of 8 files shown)


test/RelayClient.test.ts
```diff
@@ -1578,81 +1592,56 @@ describe('RelayClient', function () {
...
-        await expect(verifyWithRelayHub).to.be.rejectedWith(error.message);
-      });
-
-      it('should throw if enveloping request will fail due to invalid gas price', async function () {
-        const error = ethersLogger.makeError(
-          'Invalid gas price',
-          errors.CALL_EXCEPTION
-        );
-        const callStub = sandbox.stub().throws(error);
+      it('should throw if enveloping request will fail due to gas limite is exceeded', async function () {
```
antomor, line 1599:
```suggestion
      it('should throw if enveloping request will fail due to gas limit exceeded', async function () {
```

test/RelayClient.test.ts
```diff
@@ -1578,81 +1592,56 @@ describe('RelayClient', function () {
...
-          relayWorkerAddress,
-          FAKE_RELAY_TRANSACTION_REQUEST,
-          fakeMaxPossibleGas
-        );
-
-        expect(callStub).to.be.calledOnce;
-        await expect(verifyWithRelayHub).to.be.rejectedWith(error.message);
-      });
-
-      it('should throw if enveloping request will fail due to invalid gas price', async function () {
```
antomor, line 1607: Why did we remove this test?

test/RelayClient.test.ts
```diff
@@ -1578,81 +1592,56 @@ describe('RelayClient', function () {
           fakeMaxPossibleGas
         );
 
-        expect(callStub).to.be.calledOnce;
         await expect(verifyWithRelayHub).to.be.fulfilled;
+        expect(transactionStub).to.be.calledOnce;
       });
 
-      it('should throw if enveloping request will fail due to not enabled worker', async function () {
```
antomor, line 1585: Why did we remove this test?

src/utils.ts
```diff
@@ -296,6 +306,55 @@ const useEnveloping = (
...
+
+      return { maxPossibleGas: maxPossibleGasAttempt, transactionResult };
+    } catch (e) {
+      const error = e as Error;
+      if (!error.message.includes('Not enough gas left')) {
+        throw error;
+      }
+      maxPossibleGasAttempt = applyFactor(gasLimit, FACTOR * i);
+    }
+  }
```
antomor, line 342: I think that this logic is different from what we had before. Before the factory was increased by doubling it for each iteration, while now it grows linearly.
Furthermore, with this loop, the first two iterations we try the same value, don't we?

test/utils.ts
```diff
@@ -1,5 +1,7 @@
-import crypto from 'crypto';
+import { BigNumber, Wallet } from 'ethers';
 
-export const createRandomAddress = () =>
-  `0x${crypto.randomBytes(20).toString('hex')}`;
-export const createRandomeValue = (value: number) => `${Math.random() * value}`;
+export const createRandomAddress = () => Wallet.createRandom().address;
+export const createRandomValue = (value: number) =>
+  `${(Math.random() * value).toFixed(0)}`;
+export const createRandomBigNumber = (base: number) =>
```
antomor, line 6: Let's use the same argument name in these functions.

antomor (changes requested):
@franciscotobar thank you for the changes, I just have some doubts about the loop where we calculate different maxPossibleGas

src/utils.ts
franciscotobar: I can move it back to the previous logic but it accomplish almost the same, and I think that is simpler, this does not affect the logic.  
previous version:
.25*2=.5
.5*2=1
1*2=2

new version:
.25*1=.25
.25*2=.5
.25*3=.75
.25*4=1

test/RelayClient.test.ts
franciscotobar: This tests were included before since the method interacted directly with the contract, but now there is an extra layer that does it, I could include it again.

test/RelayClient.test.ts
franciscotobar: This tests were included before since the method interacted directly with the contract, but now there is an extra layer that does it, I could include it again.

test/RelayClient.test.ts
franciscotobar: updated

test/RelayClient.test.ts
franciscotobar: updated

test/RelayClient.test.ts
franciscotobar: updated

src/utils.ts
franciscotobar: updated

test/utils.ts
franciscotobar: updated

antomor (approved):
LGTM

PR #87, opened 2023-05-31
PP-778: implement the exchange cache

antomor:
## What

- It introduces a cache layer

## Why

- To avoid too many requests against the exchanges


src/pricer/utils.ts
```diff
@@ -22,35 +23,19 @@ const tokenToApi: Record<string, ExchangeApiName[]> = {
...
->();
-builders.set(
-  'coinBase',
-  (args: CoinBaseConstructorArgs = []) => new CoinBase(...args)
-);
-builders.set(
-  'coinCodex',
-  (args: CoinCodexConstructorArgs = []) => new CoinCodex(...args)
-);
+const builders = new Map<ExchangeApiName, () => ExchangeApi>();
```
antomor, line 29: Do we want to cache the value of all the exchanges?

src/pricer/utils.ts
```diff
@@ -22,35 +23,19 @@ const tokenToApi: Record<string, ExchangeApiName[]> = {
   TKN: ['testExchange'],
 };
 
+const CACHE_EXPIRATION_TIME = 60_000;
```
antomor, line 26: Should we make this parameter configurable?

jurajpiar (commented):
Left some remarks about the implementation. Also, I'm not sure that we're testing all possibilities. Some cases could include:
- `if the value is not cached` -> `should return the retrieved value`
- `if the value is not cached` -> `should perform call ...` -this one because of the state precedent. TBH I find confusing to have this state-based strategy (state being `if the value is not cached`), if no other states are tested against explicitly in the same way, like "is cached", "is cached and is expired", "is cached and is not expired".
I get that you wanted to use the beforeEach hook on those two, but I'm thinking that that might not be worth the clarity cost.

jurajpiar (commented):
Left some remarks about the implementation. Also, I'm not sure that we're testing all possibilities. Some cases could include:
- `if the value is not cached` -> `should return the retrieved value`
- `if the value is not cached` -> `should perform call ...` -this one because of the state precedent. TBH I find confusing to have this state-based strategy (state being `if the value is not cached`), if no other states are tested against explicitly in the same way, like "is cached", "is cached and is expired", "is cached and is not expired".
I get that you wanted to use the beforeEach hook on those two, but I'm thinking that that might not be worth the clarity cost.

src/pricer/utils.ts
franciscotobar: Im not sure about each exchange API require different timeouts, this cache was applied just to avoid consecutive queries in short intervals.

src/pricer/utils.ts
```diff
@@ -22,35 +23,19 @@ const tokenToApi: Record<string, ExchangeApiName[]> = {
...
-  'coinCodex',
-  (args: CoinCodexConstructorArgs = []) => new CoinCodex(...args)
-);
+const builders = new Map<ExchangeApiName, ExchangeApi>();
+builders.set('coinBase', new CoinBase());
+builders.set('coinCodex', new CoinCodex());
 builders.set(
   'coinGecko',
-  (args: CoinGeckoConstructorArgs = []) => new CoinGecko(...args)
+  new CachedExchangeApi(new CoinGecko(), CACHE_EXPIRATION_TIME)
```
franciscotobar, line 34: should the CachedExchangeApi, applied to all of the apis?

src/api/pricer/CachedExchangeApi.ts
```diff
@@ -0,0 +1,57 @@
+import log from 'loglevel';
+import type { ExchangeApi } from './BaseExchangeApi';
+import type { BigNumber as BigNumberJs } from 'bignumber.js';
+
+export type CacheValue<T> = {
```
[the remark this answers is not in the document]
antomor: I agree with you; better to make it more explicit 👍🏼

src/pricer/utils.ts
```diff
@@ -22,35 +23,19 @@ const tokenToApi: Record<string, ExchangeApiName[]> = {
...
 );
-builders.set('rdocExchange', () => new RdocExchange());
-builders.set('testExchange', () => new TestExchange());
+builders.set('rdocExchange', new RdocExchange());
+builders.set('testExchange', new TestExchange());
```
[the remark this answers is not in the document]
antomor: As discussed, it's not worth implementing a generator here since it would require many changes in other parts.

test/api/pricer/CachedExchangeApi.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  const valueFromExchange = '1';
+  let fakeExchangeAPI: ExchangeApi;
+
+  beforeEach(function () {
+    fakeExchangeAPI = {
+      queryExchangeRate: sinon.fake.resolves(new BigNumber(valueFromExchange)),
+    };
+  });
+
+  describe('if the value is not cached', function () {
```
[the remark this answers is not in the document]
antomor: Refactored to have a flat structure.

src/pricer/utils.ts
antomor: I decided to apply to all of them but to share the same hard-coded expiration time. If required, we could eventually make it configurable, but for now this is not a need.

src/pricer/utils.ts
antomor: Yes, it doesn't hurt, but I use the same expiration time for all of them. See [this comment](https://github.com/rsksmart/rif-relay-client/pull/87#discussion_r1211643386).

franciscotobar (approved):
LGTM

PR #91, opened 2024-06-06
fix/payment-gas-estimation

franciscotobar:
## What

- Fix the gas estimation process
- Introduced a new internal transaction estimated value for a new smart wallet type

## Why

- RSKj Arrowhead version have some changes in the gas cost
- New smart wallet type

### DIFF
--- a/src/RelayClient.ts
+++ b/src/RelayClient.ts
@@ -61,7 +61,6 @@ import {
   estimateInternalCallGas,
   estimatePaymentGas,
   getSmartWalletAddress,
-  isCustomSmartWalletDeployment,
   isDataEmpty,
   maxPossibleGasVerification,
   validateRelayResponse,
@@ -94,7 +93,8 @@ class RelayClient extends EnvelopingEventEmitter {
   }
 
   private _getEnvelopingRequestDetails = async (
-    envelopingRequest: UserDefinedEnvelopingRequest
+    envelopingRequest: UserDefinedEnvelopingRequest,
+    isCustom = false
   ): Promise<EnvelopingRequest> => {
     const isDeployment: boolean = isDeployRequest(
       envelopingRequest as EnvelopingRequest
@@ -133,6 +133,10 @@ class RelayClient extends EnvelopingEventEmitter {
     const value = (await envelopingRequest.request.value) ?? constants.Zero;
     const tokenAmount =
       (await envelopingRequest.request.tokenAmount) ?? constants.Zero;
+
+    if (!isCustom && this._isContractCallInvalid(to, data, value)) {
+      throw new Error('Contract execution needs data or value to be sent.');
+    }
 
     const { index } = request as UserDefinedDeployRequestBody;
 
@@ -298,23 +302,10 @@ class RelayClient extends EnvelopingEventEmitter {
       throw new Error('FeesReceiver has to be a valid non-zero address');
     }
 
-    // At this point all the properties from the envelopingRequest were validated and awaited.
-    const isCustom = await isCustomSmartWalletDeployment(envelopingRequest);
-
-    const { to, data, value } = envelopingRequest.request as {
-      to: string;
-      data: string;
-      value: BigNumberish;
-    };
-
-    if (!isCustom && this._isContractCallInvalid(to, data, value)) {
-      throw new Error('Contract execution needs data or value to be sent.');
-    }
-
     const tokenGas = await this._prepareTokenGas(
       feesReceiver,
       envelopingRequest,
-      isCustom
+      options?.isCustom
     );
 
     const updatedRelayRequest: EnvelopingRequest = {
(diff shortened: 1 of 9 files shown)


franciscotobar:
The relay-server does not need to know which smart wallet type is going to interact, but it tries to identify the type based on the parameters from the request. Before we only had two smart wallet types, but with the introduction of new types, there are some type that have the same parameters and the relay-server cannot identify it. In the current changes, I introduced a new property in the metadata from the relay request to help the relay-server, but this change can be improved by updating the relay-client, so it can be configured to update that property accordingly to its type.

src/common/relayClient.types.ts
```diff
@@ -21,7 +21,19 @@ type RequestConfig = {
...
+    RequestConfig,
+    | 'isSmartWalletDeploy'
+    | 'preDeploySWAddress'
+    | 'internalEstimationCorrection'
+    | 'estimatedGasCorrectionFactor'
+  >;
+
 //FIXME name standardization
+/**
+ * @deprecated The method was replaced by {@link PaymentGasEstimationParams}
```
antomor, line 35:
```suggestion
 * @deprecated The type was replaced by {@link PaymentGasEstimationParams}
```

src/utils.ts
```diff
@@ -362,17 +438,24 @@ const applyFactor = (value: BigNumberish, factor: number) => {
   );
 };
 
+const isDataEmpty = (data: string) => {
+  return ['', '0x', '0x00'].includes(data);
```
antomor, line 442: '0x00' is also included in the check `BigNumber.from(data).isZero()`

src/utils.ts
```diff
@@ -142,7 +218,7 @@ const getSmartWalletAddress = async ({
 
   const recovererAddress = recoverer ?? constants.AddressZero;
 
-  const logicParamsHash = data ?? SHA3_NULL_S;
+  const logicParamsHash = data ? utils.keccak256(data) : SHA3_NULL_S;
```
antomor, line 221: Good catch 👏🏽

antomor, line 221: Are we sure that `data` isn't already hashed?

src/utils.ts
```diff
@@ -69,6 +78,73 @@ const estimateInternalCallGas = async ({
...
+    relayData: { callForwarder, gasPrice, feesReceiver },
+  } = relayRequest;
+
+  if (tokenAmount.toString() === '0') {
+    return constants.Zero;
+  }
+
+  let tokenOrigin: PromiseOrValue<string> | undefined;
+
+  if (isDeployRequest(relayRequest)) {
```
antomor, line 98: Do we really need this check? It looks like we do the same before calling this function.

antomor (commented):
@francisco Thank you for this PR... I've just left some doubts, but overall the PR looks ok.

src/utils.ts
franciscotobar: Yes, since the two functions are different and can be called from different places.

src/utils.ts
franciscotobar: It could be that is already hashed but it will because the user is sending the data in a wrong way.
As you can see [here](https://github.com/rsksmart/rif-relay-contracts/blob/main/contracts/factory/CustomSmartWalletFactory.sol#L205), the data is hashed but [here](https://github.com/rsksmart/rif-relay-contracts/blob/main/contracts/factory/CustomSmartWalletFactory.sol#L250), is not, so unless that we change the contract we need to hash it.

src/utils.ts
franciscotobar: But here we are not checking data as a number.

src/common/relayClient.types.ts
franciscotobar: updated

antomor: @franciscotobar I still see it 😅

src/utils.ts
antomor: Sure, but it looks like in any part where we call it, we do the same before calling the function... I think we should change it in the future

src/RelayClient.ts
```diff
@@ -271,7 +273,7 @@ class RelayClient extends EnvelopingEventEmitter {
   ): boolean {
     return (
       to != constants.AddressZero &&
-      data === '0x00' &&
+      isDataEmpty(data.toString()) &&
       BigNumber.from(value).isZero()
```
antomor, line 277: According to [this comment](https://github.com/rsksmart/rif-relay-client/pull/91#discussion_r1667727589), we may probable remove the line `BigNumber.from(value).isZero()`.

src/common/relayClient.types.ts
franciscotobar: updated

antomor (approved):

PR #93, opened 2024-07-30
fix/estimation-payment-gas

franciscotobar:
## What

- Fix the gas estimation process

## Why

- Bad estimation process


antomor (approved):
LGTM

Issue #92, opened 2024-07-30
Implement LinearFit Estimation


kilrau:
https://github.com/rsksmart/rif-relay-client/blob/0969a115ea76deef0fee63e77189cc22c0fbd181/src/gasEstimator/utils.ts#L62

PR #96, opened 2024-09-01
estimation-no-signature

franciscotobar:
## What

- Estimation without signature

## Why

- UX improvement to not require additional signed tx

## Refs

- https://github.com/rsksmart/rif-relay-contracts/pull/168
- It solves #92

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "typescript": "4.8.2"
   },
   "peerDependencies": {
-    "@rsksmart/rif-relay-contracts": "github:rsksmart/rif-relay-contracts#estimation-no-signature",
+    "@rsksmart/rif-relay-contracts": "2.1.1-beta.1",
     "ethers": "^5.7.0"
   },
   "publishConfig": {
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "typescript": "4.8.2"
   },
   "peerDependencies": {
-    "@rsksmart/rif-relay-contracts": "^2.1.1-beta.0",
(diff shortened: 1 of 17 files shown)


franciscotobar:
We still need to decide the `version`.

package.json
```diff
@@ -90,7 +90,7 @@
     "typescript": "4.8.2"
   },
   "peerDependencies": {
-    "@rsksmart/rif-relay-contracts": "^2.1.1-beta.0",
+    "@rsksmart/rif-relay-contracts": "github:rsksmart/rif-relay-contracts#estimation-no-signature",
```
antomor, line 93: To be changed once we merge the contracts

src/utils.ts
```diff
@@ -121,7 +123,7 @@ const estimatePaymentGas = async ({
     // This may be wrong in case the feesReceiver performs some operations using
     // the msg.sender address in the fallback/receive function.
     return await estimateInternalCallGas({
-      from: feesReceiver,
+      from: tokenOrigin,
```
antomor, line 126: If we introduce this back, we'll need to change the comments to clarify that it works on rskj, while in other EVMs may not work as expected (Anvil is an example)

antomor (approved):
LGTM

antomor (approved):
LGTM

src/utils.ts
antomor: I decided to include a check against the chainID and set it to `tokenOrigin` in rskj and to `feesReceiver` in other cases

antomor:
We decided to abandon the linear fit implementation and use a different approach, taking into account the costs involved into the different stages of the transaction and including some cushions to be sure the estimation is enough for executing the transaction.

PR #97, opened 2024-09-05
fix: nonce from worker signer

franciscotobar:
## What

- Additional gas cost for touched accounts or storage

## Why

- After the owner already interacted with RIF Relay there are some storage that are touched, this storage needs cost to be subtracted from the request

### DIFF
--- a/src/gasEstimator/gasEstimator.ts
+++ b/src/gasEstimator/gasEstimator.ts
@@ -76,7 +76,7 @@ const estimateRelayMaxPossibleGas = async (
  *      the costs are included in all of them. These costs can be subtract by using the following constants:
  *      - POST_RELAY_DEPLOY_GAS_COST: transfer back cost is always included since we cannot know if the smart wallet is going
  *        to have some balance after the execution
- *      - OWNER_ALREADY_TOUCHED: Owner already touched, while we are doing the transfer back the owner may or not be a new account
+ *      - OWNER_ALREADY_TOUCHED: Owner already touched, the owner of the smart wallet was previously touched
  * @param relayRequest
  * @param signer
  * @param options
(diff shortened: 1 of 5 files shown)


src/gasEstimator/gasEstimator.ts
```diff
@@ -63,10 +68,15 @@ const estimateRelayMaxPossibleGas = async (
...
- *      to have some balance after the execution
+ *    - POST_DEPLOY_NO_EXECUTION: when there is no execution in deployment, an additional 4_000 are needed
+ *    - POST_DEPLOY_EXECUTION: when there is execution in deployment, an additional 1_500 are needed
+ *    - STORAGE_REFUND: After the nonces were already touched by a smart wallet deployment or smart wallet execution, we need to subtract 15_000
+ * - Manually subtract:
+ *    - Without signature, we cannot analyze if there some of the gas costs should be included or not, to avoid underestimating the relay transaction
+ *      the costs are included in all of them. These costs can be subtract by using the following constants:
+ *      - POST_RELAY_DEPLOY_GAS_COST: transfer back cost is always included since we cannot know if the smart wallet is going
+ *        to have some balance after the execution
+ *      - OWNER_ALREADY_TOUCHED: Owner already touched, the owner of the smart wallet was previously touched
```
antomor, line 79: I'd rename to something like `ADDRESS_EXISTS` or `ACCOUNT_ALREADY_CREATED`

src/gasEstimator/utils.ts
```diff
@@ -76,10 +79,22 @@ const resolveSmartWalletAddress = async (
     : callForwarder;
 };
 
+const touchedAccount = async (address: string) => {
```
antomor, line 82: Something like:
```suggestion
const isAccountCreated = async (address: string) => {
```

antomor (approved):
LGTM

